PR #1, opened 2026-05-10
First stab at intro, use-cases, and some privacy properties

SamuelSchlesinger:
Was trying to just do intro and use cases, but ended up feeling incomplete without some of the privacy properties.


thibmeu (approved):
created #3 with comments. good to go

PR #4, opened 2026-05-12
Adjust Intro and add Endorsement as a term

dennisjackson:
I've made a pass on the intro. Mostly to make the information flows a bit more explicit and the distinction with Privacy Pass a bit sharper. 

Given we're using Endorsements in the name, something in the architecture should be called 'Endorsement'. I think it makes sense for Anchors to issue Endorsements and Moderators to issue Credentials.

### DIFF
--- a/draft-schlesinger-mole-architecture.md
+++ b/draft-schlesinger-mole-architecture.md
@@ -71,20 +71,20 @@ the user maintains an account, made a payment or provided some other scare
 signal to the Origin. MoLE enables such Origins to act as Anchors which issue
 Endorsements to users, suitable for bootstrapping trust in other contexts.
 
-MoLE enables multiple Origins to share an access policy through a new party, a
-Moderator. The Moderator holds a list of trusted Anchors, and can use their
+MoLE enables multiple Origins to share an access policy through a
+Moderator. The Moderator holds a list of trusted Anchors, and can use the Anchor's
 Endorsements to issue Credentials to Users which are suitable for use on any of
 it's associated Origins. The Moderator can adjust their confidence in the
 credentials presented by a user over time. This allows the Moderator to reduce
-friction for well-behaved users and increase friction for users which are
+friction for honest users and increase friction for users which are judged to be
 misbehaving (e.g. spamming, credential stuffing, etc).
 
-Building such a system without privacy would be trivial, but unacceptable. MoLE
-delivers strong privacy protections for users. Origins and Moderators cannot
+MoLE
+delivers this functionality whilst providing strong privacy protections for users. Origins and Moderators cannot
 learn which Anchors a user has access to, only that it holds at least one
 suitable Endorsement from the Moderator's trusted list (Anchor Blindness).
 Further, even if Origins, Moderators and Anchors all collude in an attempt to
-violate User Privacy, the user's presentation of their credentials and
+violate User privacy, the user's presentation of their credentials and
 endorsements cannot be linked across different contexts (Unlinkability).
 
 MoLE is inspired by the Privacy Pass architecture {{RFC9576}}, but differs in
@@ -103,7 +103,7 @@ for deployment in contexts like the Web, requires stronger privacy properties
 than are delivered by Privacy Pass. For example, MoLE cannot rely on
 non-collusion assumptions between Issuer and Attester as is required in Privacy
 Pass's rate-limiting deployments and must instead ensure it's privacy properties
-are founded on suitable Post-Quantum cryptographic assumptions.
+are founded on suitable cryptographic assumptions.
 
 This document specifies MoLE roles, privacy and security requirements, and
 deployment considerations. Wire protocols and cryptographic instantiations are
--- a/draft-schlesinger-mole-architecture.md
+++ b/draft-schlesinger-mole-architecture.md
@@ -238,7 +238,9 @@ The following terms are used throughout this document:
 **Policy:**
 : Rules used by a Moderator or Origin to evaluate presentations.
 
-# Architecture overview {#architecture}
+# Architecture {#architecture}
+
+## Overview
 
 The Client obtains a credential from an Anchor, presents it to a Moderator, and
 uses the resulting Moderator credential when sending requests to an Origin.
@@ -270,8 +272,13 @@ CredentialFinalization     |                        |                       |
 ~~~
 {: #fig-mole-architecture title="MoLE Architecture"}
 
+## Protocols
 
+### Endorsement
 
+### Credential Issuance
+
+### Credential Presentation
 
 # Privacy Properties {#privacy-properties}
 
@@ -343,6 +350,7 @@ These channels need deployment-specific mitigations. For example, {{OBLIVIOUS-HT
 can hide network metadata between the Client and Anchor or
 Moderator. The Client-Origin channel is outside this architecture.
 
+# Deployment Considerations
 
 # Security Considerations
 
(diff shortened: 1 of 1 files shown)


draft-schlesinger-mole-architecture.md
```diff
@@ -189,21 +212,24 @@ architecture document but may be included in companion documents.
...
 **Client:**
-: An entity that seeks authorization to an Origin.
+: An entity that seeks access to resources held by an Origin.
 
 **Origin:**
 : An entity that consumes presentations from Clients and uses them to make
   authorization decisions.
 
+**Anchor:**
+: An entity that issues Endorsements to Clients based on scarce signals.
```
thibmeu, line 222: to confirm: this means that the endorsement can take a lot of forms, not only a credential?

thinking: blindrsa, signatures, token, etc

draft-schlesinger-mole-architecture.md
```diff
@@ -189,21 +212,24 @@ architecture document but may be included in companion documents.
...
+
+**Endorsement:**
+: A cryptographic object issued by an Anchor to a Client.
```
thibmeu, line 225: is cryptographic made to mean it's verifiable by another party? something else?

draft-schlesinger-mole-architecture.md
```diff
@@ -62,25 +62,48 @@ privacy preferences and users with accessibility needs. {{INTERNET-END-USER}}
...
+MoLE aims to reduce this friction, whilst maintaining the privacy of users, by
+enabling new information flows between Origins subject to cryptographically
+enforced limits.
+
+Some Origins have access to relatively rich context about a user, e.g. because
+the user maintains an account, made a payment or provided some other scare
+signal to the Origin. MoLE enables such Origins to act as Anchors which issue
+Endorsements to users, suitable for bootstrapping trust in other contexts.
+
+MoLE enables multiple Origins to share an access policy through a new party, a
```
thibmeu: do we have to mention it's a new party? maybe "a party we introduce and call a Moderator" instead

```suggestion
MoLE enables multiple Origins to share an access policy through a
```

draft-schlesinger-mole-architecture.md
```diff
@@ -62,25 +62,48 @@ privacy preferences and users with accessibility needs. {{INTERNET-END-USER}}
...
+Moderator. The Moderator holds a list of trusted Anchors, and can use their
```
thibmeu: I had to re-read `their` and infer it came from the Anchor based on "Endorsement". better write it directly

```suggestion
Moderator. The Moderator holds a list of trusted Anchors, and can use the Anchor's
```

draft-schlesinger-mole-architecture.md
```diff
@@ -62,25 +62,48 @@ privacy preferences and users with accessibility needs. {{INTERNET-END-USER}}
...
+Endorsements to issue Credentials to Users which are suitable for use on any of
+it's associated Origins. The Moderator can adjust their confidence in the
+credentials presented by a user over time. This allows the Moderator to reduce
+friction for well-behaved users and increase friction for users which are
```
thibmeu: personal bias: well-behave is a judgement from the server, while honest is the way a client perceive it's own behaviour. I think we want the later

```suggestion
friction for honest users and increase friction for users which are
```

draft-schlesinger-mole-architecture.md
```diff
@@ -62,25 +62,48 @@ privacy preferences and users with accessibility needs. {{INTERNET-END-USER}}
...
+misbehaving (e.g. spamming, credential stuffing, etc).
```
thibmeu, line 80: "which they perceived as misbehaving" instead of simply "misbehaving"?

draft-schlesinger-mole-architecture.md
```diff
@@ -62,25 +62,48 @@ privacy preferences and users with accessibility needs. {{INTERNET-END-USER}}
...
+
+Building such a system without privacy would be trivial, but unacceptable. MoLE
```
thibmeu: i don't think the sentence adds a lot

i'm also not sure it's unacceptable if that's the only thing you can deploy and that addresses your problem.
there are enfvironment for which it's actupally required to have it without privacy

however, privacy is desireable for an open web setup.

```suggestion
MoLE
```

draft-schlesinger-mole-architecture.md
```diff
@@ -62,25 +62,48 @@ privacy preferences and users with accessibility needs. {{INTERNET-END-USER}}
...
+delivers strong privacy protections for users. Origins and Moderators cannot
+learn which Anchors a user has access to, only that it holds at least one
+suitable Endorsement from the Moderator's trusted list (Anchor Blindness).
+Further, even if Origins, Moderators and Anchors all collude in an attempt to
+violate User Privacy, the user's presentation of their credentials and
```
thibmeu: is capitalisation on purpose?

```suggestion
violate User's privacy, the user's presentation of their credentials and
```

draft-schlesinger-mole-architecture.md
```diff
@@ -62,25 +62,48 @@ privacy preferences and users with accessibility needs. {{INTERNET-END-USER}}
...
-can carry policy state across presentations. These differences motivate a
-separate architecture.
+a number of aspects.
+
+Firstly, MoLE provides greater utility to Origins by enabling Moderators to
+dynamically adjust access in response to how a credential is used. Privacy Pass
+enforces a flat rate limit based on access to an underlying credential which
+cannot be curtailed even if usage is obviously abusive. This leads to
+fundamentally different information flows, privacy analysis and cryptographic
+techniques.
```
thibmeu, line 98: not sure how to correctly phrase it, but privacy pass could allow for what we do in mole.
what privacy pass does not define is state management, which we need, nor anchor blindness, which we could define as well.

i feel we could say that privacy pass does not cover state management, and does not have built in primitive for blindness

draft-schlesinger-mole-architecture.md
```diff
@@ -62,25 +62,48 @@ privacy preferences and users with accessibility needs. {{INTERNET-END-USER}}
...
+
+Secondly, MoLE targets a deployment in an open ecosystem where multiple Anchors,
+Moderators and Origins coexist with different policies. This openness, necessary
+for deployment in contexts like the Web, requires stronger privacy properties
+than are delivered by Privacy Pass. For example, MoLE cannot rely on
+non-collusion assumptions between Issuer and Attester as is required in Privacy
+Pass's rate-limiting deployments and must instead ensure it's privacy properties
+are founded on suitable Post-Quantum cryptographic assumptions.
```
thibmeu: feels that PQ is namedropped here. i think this is something that can be left to the security and privaqcyn considerations sections, that all scheme must consider PQ safety

thibmeu (approved):
the split between endorsment and credential is great. the added text in the intro adds a bit of precision

only thing to fix is the reference to privacy pass which now seems incorrect

draft-schlesinger-mole-architecture.md
dennisjackson: Hm. I think I would anchor the distinction around the difference in outcome rather than mechanic used (stateful credential, blind presentation). For example ARC is arguably a stateful privacy pass credential (since the client must update their counter). Agree it would be good to sharpen this text further though.

draft-schlesinger-mole-architecture.md
dennisjackson: I don't think that's implied by this text but I agree it would be reasonable to have an IANA registry.

draft-schlesinger-mole-architecture.md
dennisjackson: I don't think that's implied either way. I used this for commonality with the text for credential. I would interpret it to mean just that it's some bytes used in cryptographic algorithms, a pretty hazy concept. Improvements welcome!

PR #5, opened 2026-05-12
Section Headings

dennisjackson:


draft-schlesinger-mole-architecture.md
```diff
@@ -270,8 +272,13 @@ CredentialFinalization     |                        |                       |
 ~~~
 {: #fig-mole-architecture title="MoLE Architecture"}
 
+## Protocols
```
thibmeu, line 275: is this protocols or deployments?

i favour the later as I don't think we'll dive into protocols in here, but I may be wrong

thibmeu (approved):
lgtm

not sure we'll use the protocol section, or if we'll rename it, but that's needed down the line

draft-schlesinger-mole-architecture.md
dennisjackson: I would use this section to describe the information flow of each protocol, what it achieves and assumes, but not how it's achieved. I would say Deployments might be a different section where we could walk through how sites might deploy and use it as a solution?

PR #13, opened 2026-05-29
HTTP Transport initial

thibmeu:
This PR adds a first base for the HTTP Transport draft. It defines

1. Presentation language (TLS + QUIC extension)
2. HTTP Authentication Scheme to convey endorsement and credential presentation
3. Configuration placeholder, for Anchor and Moderator, as well as a bigger todo for anchor-set
4. Endorsement and credential registries, withoput parameters for now, to hold protocols for endorsement and credential respectively
5. Multiple TODOs that should be moved to issues

A change that's likely to happen is that the "HTTP protocol" sectino is going to be its own draft, where we can define protocol to be registered in endorsement and credential registry


draft-schlesinger-mole-architecture.md
```diff
@@ -353,6 +356,19 @@ Credential updates have to be applied before access to resources that
...
 
+### Discovery and consistency
+
+A Site challenge has to give Clients enough information to discover the
+Moderator and the policy context for the presentation. The Moderator then gives
+Clients enough information to discover the accepted Anchor material. The
+discovery method is deployment-specific. It can use URIs, lookup keys, or other
+opaque locators for out-of-band material that Clients are expected to resolve.
+
+Deployments should make both pieces of discovery material consistent across
```
wbl, line 367: How does this interact with wanting A/B testing and subresources?

draft-schlesinger-mole-http-transport.md
```diff
@@ -149,9 +149,15 @@ TODO(thibault)
 
 # Error Handling
 
-TODO: define error status codes, authentication error parameters, and recovery
-behavior for malformed messages, unsupported protocol types, failed validation,
-and unavailable Anchors or Moderators.
+Sites can use MoLE as optional authentication. A `200` response MAY include a
+`WWW-Authenticate: Mole` challenge. This tells the Client that a later request
```
wbl, line 149: This will need fleshing out to say where authenticate: Mole can be applied later.

draft-schlesinger-mole-http-transport.md
```diff
@@ -233,29 +254,45 @@ Authorization: Mole endorsement="..."
...
+In `MolePresentationContext`, `policy_context` is generated by the Moderator. It
+identifies the Moderator policy and accepted Anchor set used for this Site.
+Different sites MAY use different policy contexts while sharing the same
+Moderator. The value MUST be non-empty.
 
-The `presentation_context` field is chosen by the Site. It binds the
-presentation to a request, session, or time window. A non-empty value affects
-credential caching and replay handling.
+The `site_context` field is chosen by the Site. It binds the presentation to a
+request, session, or time window. A non-empty value affects credential caching
```
wbl, line 285: I'm not sure what credential caching means/do we need it, or can the site set a cookie?

wbl:
We can always merge and argue later

draft-schlesinger-mole-architecture.md
thibmeu: probably something that we should discuss. I feel that a/b and subresources are still satisfiable with this constraint, this has to be considered by protocols though

do you have places where such a constraint would be challenging to design a protocol against?

draft-schlesinger-mole-http-transport.md
thibmeu: agree. this is a first pass

draft-schlesinger-mole-http-transport.md
thibmeu: both? I meant this as presentation/session reuse by the client, but the site can also set a cookie after accepting the presentation. I’m not sure yet how much of that needs to be specified here.

optimising the presentation so that it does not happen too often (due to cost) is something that we should discuss in the drafts

draft-schlesinger-mole-architecture.md
```diff
@@ -353,6 +356,19 @@ Credential updates have to be applied before access to resources that
 an origin may gate or rate limit, so that Clients do not simply ignore
 the update request after getting them.
 
+### Discovery and consistency
+
+A Site challenge has to give Clients enough information to discover the
+Moderator and the policy context for the presentation. The Moderator then gives
+Clients enough information to discover the accepted Anchor material. The
+discovery method is deployment-specific. It can use URIs, lookup keys, or other
```
dennisjackson, line 364: > The discovery method is deployment-specific.

But still standardized right?

draft-schlesinger-mole-http-transport.md
```diff
@@ -71,10 +132,295 @@ We target browser first. We'll need some HTTP transport. We need to iterate to h
...
+1. if protocol section goes into its own draft, this should follow
+2. I suspect we'll use JSON here. If we use key material, I think re-using JWKS is important (even if there are feelings).
+
+Anchors and Moderators publish configuration used by Clients and Sites before
+running the HTTP authentication flows. Configuration includes endpoint URLs,
+supported protocol types, public key material, and the Anchor set associated
+with each Moderator policy.
+
+TODO(thibault)
+1. Anchor-Set format. I'm not sure we can find a shared format, or if this should
```
dennisjackson: I think in practice this is going to be nearly 100% determined by the crypto used in the Endorsement protocol right?

draft-schlesinger-mole-architecture.md
thibmeu: i don't think that's something we can mandate. there are likely to be standardised discovery method, but there might be some which are deployment specific. that's not something we can prevent

draft-schlesinger-mole-http-transport.md
thibmeu: yes

draft-schlesinger-mole-http-transport.md
```diff
@@ -71,10 +132,295 @@ We target browser first. We'll need some HTTP transport. We need to iterate to h
...
+friction.
+
+A `401` response with a `WWW-Authenticate: Mole` challenge means that the Site
+requires MoLE, or another acceptable authentication scheme, before serving the
+resource. A `403` response means that the Site understood the presented MoLE
+material but did not accept it under policy.
+
+## Endorsement
+
+The attestation method is left to the Anchor. This document only defines the
```
dennisjackson: > attestation method

Nit: Attestation is such an overloaded term in this context. It's a bit wordier, but I would prefer saying something like 'The criteria for issuing an endorsement is left to the Anchor'.

dennisjackson (approved):
Left a few comments but this looks good to me! I like the opaque flows - seems like this layer should be as un opinionated as possible and we can keep the complexity in the Crypto / WebAPI layers

draft-schlesinger-mole-http-transport.md
thibmeu: that's a better wording. adding that

thibmeu: actually used "providing" instead of "issuing"

draft-schlesinger-mole-http-transport.md
thibmeu: i've removed that todo

PR #18, opened 2026-06-30
Rough work: Filling out the architecture doc

dennisjackson:
This is a first pass on filling out the rest of the architecture draft. It's still pretty rough, but throwing it up for visibility. Not too attached to any of the text, so feel free to provide early comments on anything that looks totally wrong or unnecessary.

PR #20, opened 2026-07-01
Fix some typos

cryptphil:
This PR fixes three small typos.


thibmeu (commented):
i think it's safe to wait for #18 which also fixes these issue. that's a good change though

thibmeu:
addressed with #18

PR #21, opened 2026-07-04
Terminology Fix for Endorsements and Credentials 

dennisjackson:
Previously the terminology was that Issuance and Presentation were both flows, but also Endorsements could be issued and presented which made some of the writing very confusing. 

This PR changes the terminology (in bold) so that: 

 * Endorsements are **granted** and **redeemed**. 
 * Credentials are issued and presented.
 * The flows are Endorsement, **Redemption**, Presentation.

### DIFF
--- a/draft-schlesinger-mole-architecture.md
+++ b/draft-schlesinger-mole-architecture.md
@@ -87,9 +87,9 @@ Clients that a Moderator deems untrustworthy; the Moderator can mitigate this
 either by withdrawing trust in those specific Clients or by withdrawing trust in
 the Anchor entirely.
 
-At a high level, MoLE relies on three distinct flows: Endorsement, Redemption, and
+At a high level, MoLE relies on three distinct flows: Endorsement, Redeem & Issue, and
 Presentation. In Endorsement, an Anchor grants a Client an Endorsement, which
-conveys the Anchor's trust in the Client to other parties. In Redemption, a Client
+conveys the Anchor's trust in the Client to other parties. In Redeem & Issue, a Client
 redeems an Endorsement at a Moderator and receives a Credential, allowing the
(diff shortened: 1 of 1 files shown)


thibmeu:
Grant and redeem are indeed to better identify the anchor flow
This leaves credentials with issue and presentation.

However, I'm unsure about redemption. Maybe "redemption+issuance (RI step)" given both are done at the same time?

dennisjackson:
> Grant and redeem are indeed to better identify the anchor flow This leaves credentials with issue and presentation.
> 
> However, I'm unsure about redemption. Maybe "redemption+issuance (RI step)" given both are done at the same time?

Rebased on top of your changes. I'm ok with something like "Redeem  & Issue". Other idea I have is 'Transfer flow' to keep it to one idea.

draft-schlesinger-mole-architecture.md
```diff
@@ -192,31 +192,43 @@ The following terms are used throughout this document:
...
+  has an associated state that is updated during presentation.
 
 **Predicate:**
 : A boolean test that a Moderator evaluates against a Credential's hidden state
   during a Presentation. The Moderator learns only whether the state satisfies
   the Predicate, not the state itself.
 
-**Presentation:**
-: The mechanism by which a Client proves possession of a Credential whose state
+**Endorsement Flow:**
```
thibmeu, line 210: nits for later: we could reference the dedicated sectoins

thibmeu (approved):
I like "Redeem and Issue". "Transfer" is also nice. Alternatives are "Exchange", or "Swap". Having two words is not too bad for now, given there are indeed two substep.


Approving the PR as it mostly work. If we keep redemption, it's worth openning a new issue to discuss and iterate on the naming

PR #23, opened 2026-07-05
protocols: specify the endorsement and credential protocols

thibmeu:
> [!WARNING]
> This change is rather large, because the drafts were mostly empty. I think having a high level about the compositions, then individual protocols (endorsement/credential) would be useful. If that helps, happy to push this commit as its own branch on the main github repository, so that it is rendered by github actions. The PR is meant to provide a more concrete set of discussion, and text mentions that the set of protocols is subject to change at the time of writting.

Start specifying protocols to give a more concrete hook of MoLE.

This commit adds:
1. One section per protocol with its type value, configuration, messages, and behavior,
2. Two endorsement protocols (IHAT, a placeholder name pending the crypto draft, and Longfellow over legacy credentials),
3. Three credential protocols (ACT, Privacy Pass with a reverse flow, budget Privacy Pass construction).

Design decisions worth knowing about:

- Generic envelopes, per-type structures. Every endorsement type defines `Challenge` and `Presentation`. Every credential type defines `Challenge`, `IssuanceRequest`, `IssuanceResponse`, `PresentationAndUpdate`, and `Update`. The http-transport have been updated accordingly (as that draft defined these names initially).
- Redeem & Issue is an authorization, so it uses the Mole auth scheme like presentation does. `Mole-Credential` is the single response header: `response=` for issuance, `update=` after a presentation. Is this the right move: TBD,
- Anchor interactions are left to the protocols draft, which uses POST,
- Challenge binding is clarified. It hashes the decoded challenge octets, and not the base64url ASCII header,
- Clients grease with reserved `0x?A?A` credential types so Moderators keep handling unknown types and absent presentations,
- Two registries: one for endorsement, one for credentials,
- Remove `Site` from the HTTP transport draft.

Key rotation and discovery, privacy considerations, and security considerations are mostly left as TODO.

There is a full example in the appendix.

The crypto bits that were in the previous protocol draft have been split to a `-crypto` draft.

Note: I still have not added authors beyond myself to the `http-transport` and `-protocols` draft, mostly to request consent. @dennisjackson @SamuelSchlesinger you can just say yes and we'll do that before any publication

### DIFF
--- a/draft-jms-mole-protocols.md
+++ b/draft-jms-mole-protocols.md
@@ -20,9 +20,17 @@ venue:
 #  mail: "public-antifraud@w3.org"
 #  arch: "https://lists.w3.org/Archives/Public/public-antifraud/"
   github: "Moderation-of-unLinkable-Endorsements/internet-drafts"
-  latest: "https://moderation-of-unlinkable-endorsements.github.io/internet-drafts/draft-schlesinger-mole-protocols.html"
+  latest: "https://moderation-of-unlinkable-endorsements.github.io/internet-drafts/draft-jms-mole-protocols.html"
 
 author:
+ -
+    fullname: Samuel Schlesinger
+    organization: Google LLC
+    email: sgschlesinger@gmail.com
+ -
+    fullname: Dennis Jackson
+    organization: Mozilla
+    email: ietf@dennis-jackson.uk
  -
     fullname: Thibault Meunier
     organization: Cloudflare
@@ -32,13 +40,13 @@ normative:
   ACT: I-D.draft-schlesinger-cfrg-act
   ARCHITECTURE: #I-D.draft-schlesinger-mole-architecture
     title: MoLE Architecture
-    target: https://moderation-of-unlinkable-endorsements.github.io/internet-drafts/draft-schlesinger-mole-architecture.html
+    target: https://moderation-of-unlinkable-endorsements.github.io/internet-drafts/draft-jms-mole-architecture.html
   CRYPTO: #I-D.draft-authors-mole-crypto
     title: MoLE Cryptography
     target: https://moderation-of-unlinkable-endorsements.github.io/internet-drafts/draft-authors-mole-crypto.html
   HTTP-TRANSPORT: #I-D.draft-schlesinger-mole-http-transport
     title: MoLE HTTP Transport
-    target: https://moderation-of-unlinkable-endorsements.github.io/internet-drafts/draft-schlesinger-mole-http-transport.html
+    target: https://moderation-of-unlinkable-endorsements.github.io/internet-drafts/draft-jms-mole-http-transport.html
   LONGFELLOW: I-D.draft-google-cfrg-libzk
   PRIVACYPASS-AUTH: RFC9577
   PRIVACYPASS-BATCHED: I-D.draft-ietf-privacypass-batched-tokens
--- a/draft-jms-mole-protocols.md
+++ b/draft-jms-mole-protocols.md
@@ -3,7 +3,7 @@ title: "MoLE Protocols"
 abbrev: "MoLE Protocols"
 category: info
 
-docname: draft-schlesinger-mole-protocols-latest
+docname: draft-jms-mole-protocols-latest
 submissiontype: IETF
 number:
 date:
--- a/draft-schlesinger-mole-protocols.md
+++ b/draft-schlesinger-mole-protocols.md
@@ -559,8 +559,8 @@ Moderator public key. Anything else partitions Clients and leaks state.
 
 ### Presentation and Update
 
-`Challenge` is empty for this type. Binding comes from the `Token`
-structure below.
+`Challenge` is empty for this type. The challenge octets, and therefore
+`challenge_digest`, are constant for a given Moderator configuration.
 
 ~~~ tls-presentation
 struct {
@@ -574,9 +574,13 @@ struct {
 ~~~
 
 The `token` field carries a `Token` as defined in {{PRIVACYPASS-AUTH}}.
-Challenge binding comes from the `Token` structure itself: its
-`challenge_digest` field MUST equal the digest of the Moderator's
-challenge ({{challenge-binding}}).
+Its `challenge_digest` field is fixed when the token is issued, one
+exchange before it is presented, and MUST equal the Moderator's constant
+`challenge_digest` ({{challenge-binding}}). This binds the token to the
+Moderator, not to the exchange that presents it. Anti-replay therefore
+does not come from challenge binding: it comes from the token being
+single-use. The Moderator MUST reject a token whose nonce it has already
+seen.
 
 If the Moderator's policy allows continued access, it returns an `Update`.
 If not, it returns an empty update and the Client is out of credentials.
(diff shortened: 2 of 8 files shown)


draft-schlesinger-mole-protocols.md
```diff
@@ -24,164 +24,871 @@ venue:
...
+
+~~~ aasvg
++--------+     +--------+     +-----------+
+| Client |     | Anchor |     | Moderator |
++---+----+     +---+----+     +-----+-----+
+    |              |                |
+    +--------------|--- request --->|
+    |<-------------|-- challenge ---+
+    +- exchange 1 >|                |
+    |< response 1 -+                |
```
martinthomson: Very short lines like this cause aasvg to not detect a line.  You need at least two line segments.  Make this one wider.

thibmeu: indeed. updated.

draft-schlesinger-mole-protocols.md
```diff
@@ -24,164 +24,883 @@ venue:
...
-Configuration (these bits will probably differ for MoLE, and we should define the discovery we want)
+The value 0x0000 is reserved in both registries and MUST NOT appear on the
+wire. Endorsement type 0x0001 means the Moderator establishes trust in the
+Client on its own, and no Endorsement is redeemed.
 
-1. Issuer Request URL <-- issuer name is going to be th erequest URL. Name is a footgun meant for private deployment
-2. Issuer public key
-3. Challenge value
-4. Issuer set <-- magic crypto bits?
+TODO: decide whether Moderator trust establishment (0x0001) is allowed in
```
dennisjackson: It seems useful to leave it in.

draft-jms-mole-protocols.md
```diff
@@ -24,164 +24,883 @@ venue:
...
-struct CredentialRequest {
-  uint16_t token_type = 0x0531; /* Type something(something) */
-  uint8_t truncated_token_key_id; /* Allow for multiple keys per issuer */
-  ...
-}
+Every redemption and presentation is bound to the challenge that triggered
+it. The binding value is:
+
+~~~
+challenge_digest = SHA-256(challenge)
```
dennisjackson, line 170: I'm bit confused for how this works for some of the token types. Unless there's a PRF, I don't see where to bind this in.

draft-schlesinger-mole-protocols.md
```diff
@@ -24,164 +24,883 @@ venue:
...
-## Issuance Step One: Statement and Commitment
+## Longfellow {#longfellow}
 
-Client randomly selects scalars v and gamma, and sends Yprime = v Y to the anchor.
+Endorsement type: 0x0003.
 
-The Anchor computes Zprime = x Yprime, selects three random scalars aprime, bprime, and tprime.
-It computes a commitment. It then computes T1prime = tprime Yprime, T2Prime = tprime G. It then transmits Zprime, Cprime,
-and T1prime and T2prime to the client.
+Where IHAT requires Anchors to run new cryptography, this protocol reuses
```
dennisjackson: Maybe rephrase as backwards compatible with existing practices?

draft-jms-mole-protocols.md
```diff
@@ -24,164 +24,883 @@ venue:
...
+
+### Redemption
+
+This protocol needs no type-specific challenge content: `Challenge` is
+empty. The accepted issuer set, circuit, and epoch come from configuration.
+
+The Client evaluates the circuit over its credential to produce a proof and
+a nullifier. The nullifier is derived, inside the circuit, from a
+credential-bound secret and the current epoch, so one credential yields
+exactly one valid nullifier per epoch.
```
dennisjackson, line 386: Nice, sorry didn't see this before I went and looked at the link to the circuit

draft-schlesinger-mole-protocols.md
```diff
@@ -24,164 +24,883 @@ venue:
...
+~~~ tls-presentation
+struct {
+  opaque circuit_id[32];
+  opaque nullifier<V>;
+  opaque proof<V>;
+} Presentation;
+~~~
+
+The public inputs to the proof are the accepted issuer set, the epoch, the
+nullifier, and `challenge_digest` ({{challenge-binding}}). The Moderator
```
dennisjackson: How does the challenge_digest work?

draft-schlesinger-mole-protocols.md
```diff
@@ -24,164 +24,883 @@ venue:
...
+verifies the proof using the verifier of {{LONGFELLOW}}, then applies the
+same epoch and nullifier-freshness checks as IHAT redemption.
+
+### Differences from IHAT
+
+The Anchor does not participate in MoLE and need not know it exists. As a
+consequence, the Anchor cannot limit how many Endorsements a Client
+obtains. Scarcity comes only from the one-nullifier-per-epoch rule.
+Deployments that need Anchor-controlled scarcity should prefer IHAT.
```
dennisjackson: I think it would be better to say that the Anchor should be aware. A logical place to trigger this is in the context of a website with purely client side javascript. This also means that scarcity is possible. The ability to do it via adversarial interop is worth mentioning but shouldn't be the focus IMO.

draft-schlesinger-mole-protocols.md
```diff
@@ -24,164 +24,883 @@ venue:
...
+### Limitations
+
+This protocol does not bind the update to the presented credential. Token
+issuance is blind, so a Client holding two tokens can present one and
+direct every update to the other. {{ARCHITECTURE}} requires that updates
+provably apply to the presented credential. This protocol relaxes that
+requirement in exchange for running on deployed Privacy Pass code. The
+consequence is that negative updates (revocation by withholding) only work
+against Clients that hold a single credential, which the Moderator cannot
+verify. See {{security-considerations}}.
```
dennisjackson: In the context of a single bit, does this matter? Aren't the two tokens fungible?

draft-jms-mole-protocols.md
```diff
@@ -24,164 +24,883 @@ venue:
...
+direct every update to the other. {{ARCHITECTURE}} requires that updates
+provably apply to the presented credential. This protocol relaxes that
+requirement in exchange for running on deployed Privacy Pass code. The
+consequence is that negative updates (revocation by withholding) only work
+against Clients that hold a single credential, which the Moderator cannot
+verify. See {{security-considerations}}.
+
+TODO: define a device binding mechanism, issuing tokens bound to a Client
+key so that presentation requires proof of possession. This would restore
+the binding between update and presented credential. Open problem.
```
dennisjackson, line 600: Only if you trust the client?

dennisjackson (approved):
LGTM - A few questions and clarifications below. A lot of which are to do with how the presentation binding gets used or why it's necessary.

draft-schlesinger-mole-protocols.md
thibmeu: todo removed

draft-jms-mole-protocols.md
thibmeu: yes, that's a pre-requisite here. there are different level of trust though, therefore the todo

draft-schlesinger-mole-protocols.md
thibmeu: this is part of the requirement you recently removed from the architecture that updates should be applied to the same credential. given that it's been removed, i'll remove that paragraph and leave the TODO

draft-schlesinger-mole-protocols.md
thibmeu: that would help indeed. updating the text

draft-schlesinger-mole-protocols.md
thibmeu: challenge_digest is a public input

thibmeu:
rather large rebase, apologies for all the commits

PR #36, opened 2026-07-12
[architecture] fix diagrams

thibmeu:
diagram rendering had some drawing discontinuity. this fixes them they'll likely have to be reworked down the line, but this is a minimal edit

### DIFF
--- a/draft-jms-mole-architecture.md
+++ b/draft-jms-mole-architecture.md
@@ -260,6 +260,8 @@ presently in good standing with the Moderator in order to access a resource.
     |                     |                         |
     |                     +====== Presentation ====>|
     |                     |                         |
+    |                     |<======== Update ========|
+    |                     |                         |
 ~~~
 {: #fig-mole-architecture title="MoLE Architecture Overview"}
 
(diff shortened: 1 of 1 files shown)


SamuelSchlesinger (approved):

draft-jms-mole-architecture.md
```diff
@@ -243,7 +243,6 @@ presently in good standing with the Moderator in order to access a resource.
 +---+----+            +---+----+              +-----+-----+
     |                     |                         |
     |<~~~ Interaction ~~~>|                         |
-    |                     |                         |
```
martinthomson, line 246: I really think that you need to use more vertical space in your diagrams.  The ones you usually use are too cramped.  (That's taste, of course, but it also has problems with readability.)

thibmeu: in this specific instance, I was trying to have interaction and endorsement closer to highlight that they work "in the same flow". So the overview still has the same three flows that are presented below visually.

But point taken. I'll think of other ways to do so to better use the space. I like the boxes you've used around "if needed" for instance

dennisjackson (approved):

thibmeu:
updated the commit to have slightly more space based on @martinthomson suggestion. it still includes the fixes. the diagrams may change down the line, but that's future work

PR #39, opened 2026-07-17
Add more text to try to define proof

wbl:
Still much more editorial work to do


wbl:
Closing since Deep is working on the TZ IHAT writeup

PR #44, opened 2026-08-08
vole-ihat: Add intro and overview

cjpatton:
Also, add Tianyu as editor.

### DIFF
--- a/draft-authors-mole-vole-ihat.md
+++ b/draft-authors-mole-vole-ihat.md
@@ -20,12 +20,157 @@ venue:
 author:
 
   -
+    fullname: Tianyu Zhang
+    organization: University of Illinois at Urbana-Champaign
+    email: tianyuz@illinois.edu
+
+  -
     fullname: Christopher Patton
     organization: Cloudflare
     email: chrispatton+ietf@gmail.com
 
 normative:
+  IHAT: I-D.draft-authors-mole-crypto
+  FIPS202:  # TurboShortSHAKE128(M,d) = Sponge[Keccak-p[800, 12], pad10*1, 800–256](M||1111,d)
+    title: "SHA-3 Standard: Permutation-Based Hash and Extendable-Output Functions"
+    target: https://doi.org/10.6028/NIST.FIPS.202
+    date: 2015-08
+    seriesinfo:
+      "FIPS PUB": "202"
+    author:
+      -
+        org: National Institute of Standards and Technology (NIST)
+  FAEST:
+    title: "The FAEST Signature Scheme"
+    target: https://faest.info/faest-spec-v2.0.pdf
+    date: 2025
+    seriesinfo:
+      "Version": "2.0"
+    author:
+      -
+        ins: C. Baum
+        name: Carsten Baum
+      -
+        ins: L. Braun
+        name: Lennart Braun
+      -
+        ins: W. Beullens
+        name: Ward Beullens
+      -
+        ins: C. Delpech de Saint Guilhem
+        name: Cyprien Delpech de Saint Guilhem
+      -
+        ins: M. Klooß
+        name: Michael Klooß
+      -
+        ins: C. Majenz
+        name: Christian Majenz
+      -
+        ins: S. Mukherjee
+        name: Shibam Mukherjee
+      -
+        ins: E. Orsini
+        name: Emmanuela Orsini
+      -
+        ins: S. Ramacher
+        name: Sebastian Ramacher
+      -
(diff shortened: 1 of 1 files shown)


draft-authors-mole-vole-ihat.md
```diff
@@ -36,28 +181,214 @@ Issuer-Hiding Anonymous Token (IHAT) protocol. Moussaka is based on the UOV
...
+1. The Anchor does not sign the commitment directly. Instead, it "updates" the
+   commitment by binding it to the issuance context `ctx_iss` as `t = Com(DO,
+   c, ctx_iss, r_iss)`. Finally, it samples a solution `s` to `P(s) = t` and
+   sends `s, r_iss` to the Client.
+
+1. The Client finalizes issuance by computing `t` and checking that `P(s) = t`.
+
+The key difference between this protocol and blind RSA is that the Anchor
+observes the signature `s` in plaintext. To ensure unlinkability, the Client
+must hide the signature from the Moderator during redemption.
```
tnyuzg: This should be correct. But just to confirm my understanding, in blind RSA, the issuer can see the signature it signs but the user can rerandomize the signature?

draft-authors-mole-vole-ihat.md
```diff
@@ -36,28 +181,214 @@ Issuer-Hiding Anonymous Token (IHAT) protocol. Moussaka is based on the UOV
...
+
+Let `P[1], ..., P[L]` denote the Anchor Set. To redeem a token, the Client
+presents the nullifier in plaintext along with a VOLEitH proof-of-knowledge of
+`s, r_red, s_iss` for which `P[i](s) = t` for some `i in [L]` without revealing
+`i`. In order to prove this statement with VOLEitH, we need to express it as an
+arithmetic circuit.
+
+Our arithmetization exploits the simple structure of UOV signatures. Namely,
+for any public key  `P` it holds that `P(0^n) = 0^m`. The idea is to extend the
```
tnyuzg: There are different variants of UOV. The one used in NIST has no constant terms so `P(0^n) = 0^m` holds but the very first UOV paper has constant terms.

But constant terms don't contribute to security, so ppl remove it later.

But no need to change here

draft-authors-mole-vole-ihat.md
cjpatton: Actually, the server produces a "blind signature". The client removes the blinding factor, and the result is a valid signature of the message under the public exponent.

draft-authors-mole-vole-ihat.md
cjpatton: Great call out, I'll make a note of this!

draft-authors-mole-vole-ihat.md
```diff
@@ -36,28 +181,214 @@ Issuer-Hiding Anonymous Token (IHAT) protocol. Moussaka is based on the UOV
...
+witness `s` into `z[1], ..., z[L]` for which `z[j] = s` if `j = i` and `0^n`
+otherwise. The circuit evaluates the following relation:
+
+~~~
+  P[1](z[1]) + ... + P[L](z[L]) =
+    Com(DO, Com(DI, nf, ctx_red, r_red), ctx_iss, r_iss)
+~~~
```
tnyuzg, line 317: I think maybe a safer and also efficient way is to use circuit

one-hot vec constraint: b[i]·(b[i]−1)=0,  \sum_i b[i]= 1
and
(\sum_i P[i]b[i]) z = com

cjpatton: I'm not sure I follow. What are the `b[i]`s?

tnyuzg: b is a vector where only one element is 1 and all others are 0, a selector vector

cjpatton: EDIT: Ah got it. But `P[i]b[i]` is a bit ill-defined. I suppose you mean this:

```python
Q = [0] * len(P[0])  # over F_q
for i in range(L):  # L is the number of anchor keys
  for j in range(len(P[i])):
     Q[j] += b[i] * P[i][j]
return Q(s) == com
```

cjpatton: Revised after discussion:

```python
r = [0] * n
for i in range(L):
  t = P[i](s)
  for j in range(n):
    r[j] += b[i] * t[j]
```

cjpatton: Done.

tnyuzg (approved):

PR #47, opened 2026-08-12
vole-act: Fill in Ratatouille draft with protocol content

tnyuzg:
Port the Ratatouille write-up into the draft skeleton in kramdown-rfc format: introduction, conventions, the full protocol specification (key generation, commitment/tag, issuance, and spend-and-reissue), the two commitment instantiations (Keccak-p[800,12] and the BFMRSV25a MQ commitment), and performance numbers. Reference links and internal section references are converted to {{...}} citations and cross-refs.

### DIFF
--- a/draft-authors-mole-vole-act.md
+++ b/draft-authors-mole-vole-act.md
@@ -260,7 +260,7 @@ signature scheme and the VOLE-in-the-Head proof system (a component of the
 FAEST signature scheme). The moderator issues a token by signing a commitment
 to the client's state; to present a token, the client proves knowledge of this
 signature and the opening of the commitment. Two variants of the protocol are
-specified, one in which the commitment is instantiated with Keccak-p[800,12]
+specified, one in which the commitment is instantiated with `Keccak-p[800,12]`
 and another in which a UOV-friendly commitment scheme is used instead.
 
 --- middle
@@ -333,7 +333,7 @@ kilobytes to tens of kilobytes. Nonetheless, the protocol remains practical for
 modern web services, with efficient signing, verification, and proof
 generation.
 
-Unlike ACT, Ratatouille tokens are publicly verifiable in the sense that it can be presented to any Moderator, not just the Moderator that issued it. Of course, this has implications for privacy, since the token necessarily reveals the identity of the Moderator that issued it. 
+Unlike ACT, Ratatouille tokens are publicly verifiable in the sense that it can be presented to any Moderator, not just the Moderator that issued it. Of course, this has implications for privacy, since the token necessarily reveals the identity of the Moderator that issued it.
 
 ## Building Blocks
 
@@ -858,7 +858,7 @@ blindness are inherited unchanged. They differ only in `Com`, `Tag`, the
 deployment-wide parameters derived from `dst`, and the resulting UOV dimensions
 `(n_uov, m_uov)`.
 
-## Hash commitment (Keccak-p[800, 12]) {#hash-commitment}
+## Hash commitment (`Keccak-p[800, 12]`) {#hash-commitment}
 
 Here both `Com` and `Tag` are realized by `KP800`, a TurboSHAKE
 {{!RFC9861}} sponge over the reduced-round permutation `Keccak-p[800, 12]`
@@ -1097,7 +1097,7 @@ L/8 = 30` left zero until `Tag` writes it (no reserved bytes, since `off_refund
 > security loss.
 
 > TODO Finalize parameters MQ variant (including the UOV parameters used), pending further cryptanalysis of the commitment.
-> [Remind the reader here how the current parameters were chosen.]
+> TODO Remind the reader here how the current parameters were chosen.
 
 # IANA Considerations
 
--- a/draft-authors-mole-vole-act.md
+++ b/draft-authors-mole-vole-act.md
@@ -992,8 +992,7 @@ and V2 of {{prove-spend}} are pure `F_256` arithmetic with no Keccak in-circuit.
 The realization differs from {{hash-commitment}} in two ways:
 
 1. **`Com` is algebraic.** `Com(nf, t, r) = (EmbedNullifierBalance(nf, t) +
-   F(r), G(r))` for public general (inhomogeneous) quadratic maps `F`, `G` --
-   each with a linear term -- from the `GenerateParameters` function defined
+   F(r), G(r))` for public general quadratic maps `F`, `G` from the `GenerateParameters` function defined
    below.
 2. **`Tag` is affine and homomorphic.** The MQ commitment is
    constant-additively homomorphic, so the refund is applied by a homomorphic
--- a/draft-authors-mole-vole-act.md
+++ b/draft-authors-mole-vole-act.md
@@ -284,7 +284,7 @@ client a token worth some number of credits, and the client can later spend
 those credits anonymously, in whole or in part, without the issuer being able
 to link the spend to the issuance or to any other spend by the same client.
 
-ACT {{ACT}} builds on BBS signatures over prime-order groups. Its
+ACT {{ACT}} builds on BBS MACs over prime-order groups. Its
 *unlinkability* (the issuer cannot correlate transactions) is
 information-theoretic, following from the perfectly hiding commitments, so it
 already holds against unbounded and hence quantum adversaries. Its
@@ -1107,6 +1107,7 @@ This document has no IANA actions.
 --- back
 
 # Performance {#performance}
+{:numbered="false"}
 
 The hash instantiation of {{hash-commitment}} evaluates one in-circuit
 `Keccak-p[800, 12]` permutation per issuance and three per spend (recomputing
@@ -1192,3 +1193,5 @@ of forms, not the map size, so the extra width costs little.
 
 # Acknowledgments
 {:numbered="false"}
+
+The use of Keccak-p[800,12] in this application was proposed by Bas Westerbaan.
--- a/draft-authors-mole-vole-act.md
+++ b/draft-authors-mole-vole-act.md
@@ -377,8 +377,8 @@ This document uses the following notation:
 * `P(s)`: Application of a quadratic map `P: F_q^{n_uov} -> F_q^{m_uov}` to a
   vector `s`, yielding `(s^T P_0 s, ..., s^T P_{m_uov-1} s) in F_q^{m_uov}`,
   where each `P_i` is an upper-triangular matrix over `F_q`.
-* `<t>_L`: The `L`-bit big-endian encoding of a nonnegative integer `t < 2^L`
-  as an element of `F_2^L`.
+* `<t>_L`: The `L`-byte big-endian encoding of a nonnegative integer `t < 2^(8*L)`
+  as an element of `F_256^L` (equivalently `F_2^(8*L)`).
 * `H`: A cryptographic extendable-output hash function, modeled as a random
   oracle (see {{protocol}}). Domain separation is applied by prefixing a
   distinct label and the deployment tag `dst` to each use.
@@ -390,13 +390,13 @@ The protocol uses the following primitive data types:
 * **Bit**: An element of `F_2`.
 * **Field element**: An element of `F_q` or `F_{2^lambda}`.
 * **Field vector**: A tuple of field elements, e.g., an element of `F_q^n`.
-* **Credit value**: A nonnegative integer in the range `[0, 2^L - 1]`
+* **Credit value**: A nonnegative integer in the range `[0, 2^(8*L) - 1]`
   representing an amount of credits, encoded as `<t>_L`.
-* **Nullifier**: A uniformly random bit string of length `nu`, denoted `nf`,
+* **Nullifier**: A uniformly random byte string of length `nu`, denoted `nf`,
   that identifies a token for double-spend detection. Its length is governed by
   collision resistance across all tokens issued under one key (see
   {{parameters}}).
-* **Randomness**: A uniformly random bit string of length `rho`, denoted `r`,
+* **Randomness**: A uniformly random byte string of length `rho`, denoted `r`,
   used to hide a commitment. Its length is governed by the hiding of the
   commitment scheme.
 * **ByteString**: A sequence of bytes.
@@ -419,9 +419,21 @@ From these, the protocol builds the following objects, each defined in
 
 ## VOLE-in-the-Head Proofs of Knowledge
 
-> TODO Define Prove{} and Verify{} APIs, with references to FAEST. Give a high
-> level overview of VOLEitH. Define any terms we use here, such as "small
-> VOLE", "big VOLE", and "grinding".
+We write `pf := VOLEitH.Prove(R, X, W)` to denote proving knowledge of a witness
+`W` for which the pair `(X, W)` is in the relation `R`, where `X` is the
+instance, using the VOLE-in-the-Head proof system of {{FAEST}}. We write `v :=
+VOLEitH.Verify(R, X, pf)` to denote verification of `pf` under the same relation
+and instance, where the output `v` is a bit.
+
+Relations are specified as a sequence of constraints. Each constraint is an
+equation of the instance and witness variables; unlike a linear system, each
+side of the equation may be a polynomial over `F_q` of arbitrary degree. Because
+the degree of a constraint influences the size of the proof, a relation is
+designed to balance the size of the witness against the degree of its
+constraints.
+
+> TODO Give a high level overview of VOLEitH. Define any terms we use here, such
+> as "small VOLE", "big VOLE", and "grinding".
 
 ## Cryptographic Parameters {#parameters}
 
@@ -432,15 +444,15 @@ instantiations (ciphersuites) are given in {{instantiations}}:
 Parameters:
   - lambda: Security parameter in bits (e.g., 128). Sets the big VOLE field
             F_{2^lambda} and the default for nu and rho.
-  - nu: Nullifier length in bits. Chosen so that random nullifiers do not
+  - nu: Nullifier length in bytes. Chosen so that random nullifiers do not
         collide across all tokens issued under one key (see below).
-  - rho: Commitment randomness length in bits. Chosen for commitment hiding.
+  - rho: Commitment randomness length in bytes. Chosen for commitment hiding.
   - (n_uov, m_uov): UOV dimensions -- n_uov variables, m_uov equations -- over `F_q`.
   - (tau, N, w_grind): VOLE-in-the-Head parameters -- tau VOLE instances,
             N leaves per GGM tree (with tau * log2(N) = lambda), and a
             proof-of-work grinding parameter w_grind.
-  - L: Bit width of credit values (1 <= L <= 128). The maximum credit a
-       token can hold is 2^L - 1.
+  - L: Byte width of credit values (1 <= L <= 16). The maximum credit a
+       token can hold is 2^(8*L) - 1.
   - H: Extendable-output hash function, written H(M, d) to take input M and squeeze d bytes
        of output (e.g., the KP800 sponge over Keccak-p[800, 12] for
        lambda = 128; see hash commitment). (TODO: does this belong to Parameters?)
@@ -453,14 +465,15 @@ match, so neither weakens the token's overall security ({{system-parameters}}).
 
 The lengths `nu` and `rho` are independent. The nullifier is revealed and must
 not collide across tokens issued under one key, so a birthday bound motivates
-`nu >= 2*lambda`, and a ciphersuite defaults to `nu = 2*lambda`; a compact
-ciphersuite MAY use a smaller `nu` (e.g., `192` bits in {{hash-commitment}} and `224` bits in the MQ instantiation of
+`nu >= lambda/4` bytes (`= 2*lambda` bits), and a ciphersuite defaults to `nu =
+lambda/4`; a compact ciphersuite MAY use a smaller `nu` (e.g., `24` bytes in
+{{hash-commitment}} and `28` bytes in the MQ instantiation of
 {{mq-commitment}}), trading collision margin for size. The randomness only hides
-the commitment; its length is instantiation-dependent -- `rho = lambda` for the
-hash commitment ({{hash-commitment}}) and `rho = 8*n_com` (the `F_256` input
-length, i.e. `n_com` elements of `F_256`) for the MQ commitment
-({{mq-commitment}}). Credit values are exactly `L` bits: the issuer MAY choose
-any initial amount in `[1, 2^L - 1]`, with no cap beyond `L`.
+the commitment; its length is instantiation-dependent -- `rho = lambda/8` bytes
+for the hash commitment ({{hash-commitment}}) and `rho = n_com` (the `F_256`
+input length, i.e. `n_com` elements of `F_256`) for the MQ commitment
+({{mq-commitment}}). Credit values are exactly `L` bytes: the issuer MAY choose
+any initial amount in `[1, 2^(8*L) - 1]`, with no cap beyond `L`.
 
 # Protocol Overview {#overview}
 
@@ -539,38 +552,26 @@ organization, the service, the deployment environment, and so on. For example:
 
 ## Key Generation {#key-generation}
 
-Key generation is run once by the issuer. It produces a long-lived UOV key
-pair. The routines `ExpandSK`, `ExpandPK`, and `Upper` are those of {{UOV}}
-(Section 3.2, Figure 2): `ExpandSK` expands the secret seed into the trapdoor
-(plus a public seed), `ExpandPK` expands that public seed `seed_pk` into the
-first and second quadratic-map blocks `P^(1)` and `P^(2)`, and `Upper`
-({{UOV}}, Section 3.1) derives the third block `P^(3)` from the first two and
-the trapdoor.
-
-We use the compressed form: the public key `pk` is just the public seed
-`seed_pk` and the third quadratic-map blocks `P^(3)` while the secret key `sk`
-is the secret seed `seed_sk`.
+Key generation is run once by the issuer to produce a long-lived UOV key pair.
+The routines `UOV.CompactKeyGen`, `UOV.ExpandPK`, and `UOV.ExpandSK` are those of
+{{UOV}} (Section 3.2, Figure 2): `CompactKeyGen` samples a compact key pair
+`(cpk, csk)`, `ExpandPK` expands `cpk` into the public UOV map `P`, and
+`ExpandSK` expands `csk` into the trapdoor for `P`.
 
 > TODO Ensure all algorithms are deterministic (the caller provides the initial randomness). This way we can easily generate consistent test vectors. We'll also need to plumb this through VOLEitH proving and UOV signing.
 
-~~~
+~~~~pseudocode
 KeyGen():
-  Input: None
   Output:
-    - sk: Issuer secret key (secret seed seed_sk)
-    - pk: Issuer public key (public seed seed_pk and blocks P^(3))
+    - sk: Issuer secret key, a UOV trapdoor for P
+    - pk: Issuer public key, a UOV map P from F_q^{n_uov} to F_q^{m_uov}
 
   Steps:
-    1.  seed_sk <- {0,1}^(seed_sk_len)
-    2.  (seed_pk, O) := ExpandSK(seed_sk)   // trapdoor O in F_q^(v * m_uov),
-                                           // with v = n_uov - m_uov
-    3.  {(P_i^(1), P_i^(2))}_{i in [m_uov]} := ExpandPK(seed_pk)
-    4.  for i in [m_uov]:
-    5.      P_i^(3) := Upper( - O^T P_i^(1) O - O^T P_i^(2) O )
-    6.  sk := seed_sk
-    7.  pk := (seed_pk, { P_i^(3) }_{i in [m_uov]})
-    8.  return (sk, pk)
-~~~
+    1.  (cpk, csk) := UOV.CompactKeyGen()
+    2.  pk := UOV.ExpandPK(cpk)
+    3.  sk := UOV.ExpandSK(csk)
+    4.  return (sk, pk)
+~~~~
 
 The issuer publishes `pk`. The issuer additionally maintains a set of accepted
 nullifiers for rejecting any nullifier it has already accepted.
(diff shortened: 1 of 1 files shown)


draft-authors-mole-vole-act.md
```diff
@@ -46,7 +46,7 @@ normative:
   FAEST:
     title: "The FAEST Signature Scheme"
     target: https://faest.info/faest-spec-v2.0.pdf
-    date: 2026
+    date: 2025
```
cjpatton, line 49: Where did you get this date?

draft-authors-mole-vole-act.md
```diff
@@ -170,19 +267,928 @@ and another in which a UOV-friendly commitment scheme is used instead.
...
+the Fischlin blind-signature paradigm {{Fischlin}}, in which a client sends the
+issuer a hiding commitment to a message, the issuer blindly signs the
+commitment, and the client later proves knowledge of a signature on the
+committed message. Ratatouille generalizes this from a one-time signature to a
+spendable balance: the commitment binds a nullifier together with the remaining
+credits, and each presentation additionally proves that the spend was
+subtracted correctly and the balance stays in range, with the issuer re-signing
+the updated commitment so that every spend chains into the next token.
+
+Ratatouille is inspired by PoMFRIT {{POMFRIT}}, which builds a post-quantum
```
cjpatton: nit nit :)

```suggestion
Ratatouille is inspired by PoMFRIT {{PoMFRIT}}, which builds a post-quantum
```

draft-authors-mole-vole-act.md
```diff
@@ -170,19 +267,928 @@ and another in which a UOV-friendly commitment scheme is used instead.
...
+remains, then, is to make unforgeability post-quantum as well.
+
+This document specifies Ratatouille, a **post-quantum** variant of ACT, whose
+security rests on assumptions believed to resist quantum attack: the hardness
+of structured multivariate quadratic (MQ) problems, via the Unbalanced Oil and
+Vinegar (UOV) signature scheme {{UOV}}, and the security of symmetric
+primitives (hash functions and pseudorandom generators), via the
+VOLE-in-the-Head (VOLEitH) zero-knowledge proof system {{FAEST}}. We build on
+the Fischlin blind-signature paradigm {{Fischlin}}, in which a client sends the
+issuer a hiding commitment to a message, the issuer blindly signs the
```
cjpatton: I wanna try to be careful to not overload terminology in RFC 9474.

```suggestion
issuer a hiding commitment to a message, the issuer signs the
```

draft-authors-mole-vole-act.md
```diff
@@ -170,19 +267,928 @@ and another in which a UOV-friendly commitment scheme is used instead.
...
+   unlinkability, and double-spend prevention -- are conjectured to hold against
+   adversaries equipped with a large-scale quantum computer. Security reduces to
+   the UOV and MQ problems and to the security of standard symmetric primitives.
+6. **Performance**: The protocol is practical for modern web services.
+   Presentation proofs are larger than in the classical construction -- on the
+   order of kilobytes to tens of kilobytes -- but signing, verification, and
+   proof generation remain efficient.
 
-# Conventions and Definitions
+> TODO Is public verifiability useful?
```
cjpatton: Say what you mean by public verifiability: Moderator 1 can redeem a token issued by Moderator 2, but the identity of Moderator 1 is revealed to Moderator 2.

draft-authors-mole-vole-act.md
```diff
@@ -170,19 +267,928 @@ and another in which a UOV-friendly commitment scheme is used instead.
...
+   issuer stores accepted nullifiers and rejects repeats, without learning
+   anything that links the nullifier to a client.
+5. **Post-Quantum Security**: All security properties -- unforgeability,
+   unlinkability, and double-spend prevention -- are conjectured to hold against
+   adversaries equipped with a large-scale quantum computer. Security reduces to
+   the UOV and MQ problems and to the security of standard symmetric primitives.
+6. **Performance**: The protocol is practical for modern web services.
+   Presentation proofs are larger than in the classical construction -- on the
+   order of kilobytes to tens of kilobytes -- but signing, verification, and
+   proof generation remain efficient.
```
cjpatton: Most of this section is redundant. In particular we should assume the reader is familiar with ACT.

It's worth mentioning the communication cost, but that can be a single paragraph.

draft-authors-mole-vole-act.md
```diff
@@ -170,19 +267,928 @@ and another in which a UOV-friendly commitment scheme is used instead.
...
+
+* **Rate Limiting**: Services can issue periodic credit allowances that clients
+  spend anonymously for API calls or resource access.
+* **API Credits**: API providers can sell credit packages that developers use
+  to pay for API requests without creating a detailed usage history linked to
+  their identity. This enables:
+    * Pre-paid API access without requiring credit cards for each transaction
+    * Anonymous API usage for privacy-sensitive applications
+    * Usage-based billing without tracking individual request patterns
+    * Protection against competitive analysis through usage monitoring
```
cjpatton: Likewise here, let's leave this to the ACT doc.

draft-authors-mole-vole-act.md
```diff
@@ -170,19 +267,928 @@ and another in which a UOV-friendly commitment scheme is used instead.
...
+
+## Protocol Overview {#overview}
```
cjpatton: Overview should come after Conventions and Definitions.

draft-authors-mole-vole-act.md
```diff
@@ -170,19 +267,928 @@ and another in which a UOV-friendly commitment scheme is used instead.
...
+  publicly verifiable, are built entirely from symmetric primitives, and are
+  concretely small and fast for circuits over small binary fields.
+* **Fischlin Blind Signatures {{Fischlin}}**: A blind-signature paradigm built
+  from a commitment scheme, a digital signature, and a zero-knowledge proof of
+  knowledge. Issuance signs a commitment; presentation proves knowledge of a
+  signature over a commitment. The post-quantum instantiation of this paradigm
+  using VOLEitH and a multivariate trapdoor was studied in PoMFRIT {{POMFRIT}}.
+* **Anonymous Credit Tokens {{ACT}}**: The original ACT protocol, optimized for
+  numerical values and partial spending over prime-order groups. This document
+  can be viewed as a post-quantum analogue of that protocol.
```
cjpatton: Reframe: We should mention the primitives we build on before the overview (similar to https://github.com/Moderation-of-unLinkable-Endorsements/internet-drafts/pull/44). We can delete the {{Fischlin}} or {{ACT}} bullets completely I think, as these are already discussed in the intro.

draft-authors-mole-vole-act.md
```diff
@@ -170,19 +267,928 @@ and another in which a UOV-friendly commitment scheme is used instead.
...
+  - (tau, N, w_grind): VOLE-in-the-Head parameters -- tau VOLE instances,
+            N leaves per GGM tree (with tau * log2(N) = lambda), and a
+            proof-of-work grinding parameter w_grind.
+  - L: Bit width of credit values (1 <= L <= 128). The maximum credit a
+       token can hold is 2^L - 1.
+  - H: Extendable-output hash function (e.g., the KP800 sponge over
+       Keccak-p[800, 12] for lambda = 128; see hash commitment). (TODO)
+~~~
+
+The issuer's public key is a UOV quadratic map `P: F_q^{n_uov} -> F_q^m`; the
```
cjpatton: Can rename `n_uov` to `n` or does this cause a collision? I'd prefer to have more symmetry in the UOV parameters, i.e., `n, m` or `n_uov, m_uov`.

draft-authors-mole-vole-act.md
```diff
@@ -170,19 +267,928 @@ and another in which a UOV-friendly commitment scheme is used instead.
...
+Com(nf, t, r):
+  Input:
+    - nf: Nullifier (nu bits)
+    - t: Credit value in [0, 2^L - 1]
+    - r: Randomness (rho bits)
+  Output:
+    - c: Commitment (m*8 bits)
+
+  Steps:
+    1.  return KP800(0xAC || dst || nf || <t>_L || r, m)
```
cjpatton: `KB800()` needs a normative reference. What is `m`?

draft-authors-mole-vole-act.md
```diff
@@ -170,19 +267,928 @@ and another in which a UOV-friendly commitment scheme is used instead.
...
+revealing `c`.
+
+### Client: Issuance Request {#issue-request}
+
+~~~
+IssueRequest(T_init):
+  Input:
+    - T_init in [1, 2^L - 1]: agreed initial credit balance
+  Output:
+    - request = (c, pi_init): commitment and well-formedness proof
```
cjpatton: Let's go with `pf_init` instead of `pi_init`, since `pf` (short for "proof") is slightly more meaningful. Likewise for other VOLEitH proofs.

draft-authors-mole-vole-act.md
```diff
@@ -170,19 +267,928 @@ and another in which a UOV-friendly commitment scheme is used instead.
...
+
+~~~
+Tag(c, x):
+  Input:
+    - c: Commitment (m elements in F_256)
+    - x: Refund in [0, 2^L - 1]
+  Output:
+    - y: UOV target (m elements in F_256)
+  Steps:
+    1.  return c + EmbedRefund(x)     in F_256^m   // degree-1, no hash
```
cjpatton: Ahhhh, I forgot about how lovely this trick is. It's worth leaving a TODO here to remind ourselves to think through the security reduction here. We might have to accept losing `|x|` bits of security here for the same reason as https://github.com/Moderation-of-unLinkable-Endorsements/internet-drafts/pull/44.

draft-authors-mole-vole-act.md
```diff
@@ -170,19 +267,928 @@ and another in which a UOV-friendly commitment scheme is used instead.
...
+in-circuit hash, and land in the same regime -- the residual gap is ISA and
+language (AVX C/C++ vs aarch64 Rust). Our RATA-MQ spend (prove 85 ms / verify 57
+ms) sits in the same regime once its widened-UOV map is folded with lift tables:
+the map is ~26x larger in coefficients, but the fold cost scales with the number
+of forms, not the map size, so the extra width costs little.
 
-# Security Considerations
+# Security Considerations {#security}
 
 > TODO
```
cjpatton: Let's add TODOs here for all the things that need security analysis.
```suggestion

> TODO Security reduction for both variants to (one-more-)UOV problem. As part of this, figure out if the constant-addition hommoorphism trick for the MQ variant incurs a security loss.

> TODO Finalize parameters MQ variant, pending further cryptanalysis of the commitment. [Remind the reader here how the current parameters were chosen.]
```

draft-authors-mole-vole-act.md
```diff
@@ -170,19 +267,928 @@ and another in which a UOV-friendly commitment scheme is used instead.
...
+| `Tag(c, x)` | `c: 131`, `<x>_L: L/8 = 2` | `y: m = 131` |
+| `P(s)` (UOV) | `s: n_uov = 275` | `y: m = 131` |
+
+The message `msg = EmbedNullifierBalance(nf, t)` lays out `k = 32` bytes as `nf
+(28) || <t>_L (2) || <x>_L (2)`, with the refund slot at `off_refund = nu/8 +
+L/8 = 30` left zero until `Tag` writes it (no reserved bytes, since `off_refund
++ L/8 = k`). Derived map and signature dimensions: `F: F_256^83 -> F_256^32`,
+`G: F_256^83 -> F_256^99`, and `v = n_uov - m = 144`.
+
+# Performance {#performance}
```
cjpatton: This section should be move to the appendix. Here's an example of an appendix:
https://github.com/cfrg/draft-irtf-cfrg-vdaf/blob/main/draft-irtf-cfrg-vdaf.md?plain=1#L6523

cjpatton (approved):

draft-authors-mole-vole-act.md
```diff
@@ -170,19 +267,928 @@ and another in which a UOV-friendly commitment scheme is used instead.
...
+|-----------|------------------------------------|
+| `IssueRequest` (prove) | 24 ms |
+| `IssueResponse` (verify) | 65 ms |
+| `ProveSpend` | 85 ms |
+| `VerifyAndRefund` (verify) | 57 ms |
+
+| Proof | MQ |
+|-------|----|
+| Issuance `pi_init` | 4.3 KiB |
+| Spend `pi` | 9.0 KiB |
```
cjpatton: Can you add a similar table for the hash variant?

draft-authors-mole-vole-act.md
```diff
@@ -170,19 +267,928 @@ and another in which a UOV-friendly commitment scheme is used instead.
...
+Modern web services face a fundamental tension between operational needs and
+user privacy. Services need to implement rate limiting to prevent abuse, charge
+for API usage to sustain operations, and allocate computational resources
+fairly. However, traditional approaches require tracking client identities and
+creating detailed logs of client behavior, raising significant privacy concerns
+in an era of increasing data protection awareness and regulation.
+
+The Anonymous Credit Token (ACT) protocol {{ACT}} resolves this tension by
+enabling credit-based systems without client tracking: an issuer can grant a
+client a token worth some number of credits, and the client can later spend
```
martinthomson, line 283: Not for right now, necessarily, but this notion of "credits" is something we've agreed not to fixate on.

The token carries user-specific state, that's about it.  It might be used to support a credit-like scheme, or a reputation-like scheme, or it could just encode the output of some assessment.  The key property being that the moderator sets an initial value and can modify that value, but it doesn't get to read back the state, except through (very) occasional tests it can run over state.

draft-authors-mole-vole-act.md
```diff
@@ -170,19 +267,928 @@ and another in which a UOV-friendly commitment scheme is used instead.
...
+   blindly signs the message (i.e., signs the hiding commitment of the
+   message). The result is a credit token that the issuer cannot link to future
+   spends.
+3. **Spending (Present-and-Reissue)**: To spend credits, the client reveals the
+   nullifier of its current token and proves, in zero knowledge, that (a) it
+   holds a valid signature over that token, (b) the token's balance suffices,
+   and (c) it has correctly formed a fresh commitment for the remaining
+   balance. The issuer checks the proof, checks that the nullifier has not been
+   accepted before, records it, and blindly signs the new commitment -- issuing
+   change as a new, unlinkable token.
```
martinthomson: We'll need a version of this that skips (b).  I assume that this is possible, but you'll need to explicitly include that.

draft-authors-mole-vole-act.md
```diff
@@ -170,19 +267,928 @@ and another in which a UOV-friendly commitment scheme is used instead.
...
+first and second quadratic-map blocks `P^(1)` and `P^(2)`, and `Upper`
+({{UOV}}, Section 3.1) derives the third block `P^(3)` from the first two and
+the trapdoor.
+
+We use the compressed form: the public key `pk` is just the public seed
+`seed_pk` and the third quadratic-map blocks `P^(3)` while the secret key `sk`
+is the secret seed `seed_sk`.
+
+~~~
+KeyGen():
```
martinthomson, line 583: This should take the seed as an argument, if only to aid in testing.  The user can supply their own randomness.

draft-authors-mole-vole-act.md
tnyuzg: What's the purpose of skipping (b)? Does this make ACT degenerate to a blind signature?

draft-authors-mole-vole-act.md
tnyuzg: From my inference: NIST round 2 submission update has deadline at 2025:
https://csrc.nist.gov/projects/pqc-dig-sig/standardization/workshops-and-timeline

draft-authors-mole-vole-act.md
tnyuzg: removed.
Does the overload mean "blindly sign" is used both for act and blind signature now?

cjpatton: (My comment is reallllll nit picky, maybe just ignore it.) RFC 9474 writes "blind sign" to mean the signer signs a blinded message and produces a blinded signature. Here the signer is indeed signing a blinded message (the commitment), but the signature is not "blinded".

draft-authors-mole-vole-act.md
tnyuzg: Keep "Full post-quantum security" followed with its performance cost i.e. the communication cost.
Remove other properties' descriptions

draft-authors-mole-vole-act.md
martinthomson: Oh, this is in the architecture.  (At least, it should be.) We don't want to allow arbitrary reads of the state, but we do want to allow arbitrary updates to the state.  So yes, this is akin to a blind signature, but you still reveal that the token exists and is valid.

This is OK because the moderator has previously read a valid token off, but the value needs to be hidden because there might be other (unlinkable) places where the token has been exchanged and updated as well.  So the value might no longer be valid.  The update might restore it to being valid, but that's not something that they can test often.

draft-authors-mole-vole-act.md
tnyuzg: Yeah. And just to make sure, the credential (the proof of token by the user) can reveal the UOV public key, which identifies the moderator. Is this OK in future application of ACT in MoLE?

martinthomson: For the moderator, it's OK to identify the public key.  Though we might want to have something to help mitigate the risk of that being used to identify people.  It might be necessary for the moderator to identify the key that they expect you to use on the first reveal, in case you have obtained the token from another context. 

That is, you visit a-site.example and get a token from the moderator there.  Later, you visit b-site.example and the moderator is the same.  You want to be able to use the token you have (that's the whole point of the moderator), but you don't want to reveal which specific case issued it.  The visit to a-site.example might have used a specific key for issuance that was specific to you.  So revealing the public key could be a privacy leak.  Forcing the site to identify its own key before getting the token is probably enough to deal with that.

We expect that the moderator will be able to connect multiple presentations in the same context.  It's just that initial cross-context reveal that is potentially problematic.

draft-authors-mole-vole-act.md
tnyuzg: m is used to denote the  length of output byte

draft-authors-mole-vole-act.md
tnyuzg: n_uov is used to be distinguished from n_com of MQ com

draft-authors-mole-vole-act.md
tnyuzg: Nice! The reminder is now in security section

draft-authors-mole-vole-act.md
tnyuzg: Can I understand this as the token in ACT or MoLE is more generic than credits?

martinthomson: Yeah.  Also, we don't want to people to get the wrong impression about this being a "currency".  That's a toxic subject.

draft-authors-mole-vole-act.md
cjpatton: Nice, updated on #44 as well.

draft-authors-mole-vole-act.md
cjpatton: Agreed. @tnyuzg let's leave a TODO here for now.

```suggestion
> TODO Ensure all algorithms are deterministic (the caller provides the initial randomness). This way we can easily generate consistent test vectors. We'll also need to plumb this through VOLEitH proving and UOV signing.

~~~
KeyGen():
```

draft-authors-mole-vole-act.md
```diff
@@ -170,21 +267,840 @@ and another in which a UOV-friendly commitment scheme is used instead.
...
+This post-quantum security comes at a cost in **performance**: presentation
+proofs are larger than in the classical construction -- on the order of
+kilobytes to tens of kilobytes. Nonetheless, the protocol remains practical for
+modern web services, with efficient signing, verification, and proof
+generation.
+
+> TODO: Is public verifiability of the token useful? Using public verifiability we can enable a
+> token issued by one moderator can be redeemed at another: Moderator 1 can
+> redeem a token issued by Moderator 2, at the cost of revealing the identity of
+> Moderator 1 to Moderator 2. Is this trade-off worth supporting?
```
cjpatton: From https://github.com/Moderation-of-unLinkable-Endorsements/internet-drafts/pull/47#discussion_r3771147790 it sounds like we might actually have support for this.
```suggestion
Unlike ACT, Ratatouille tokens are publicly verifiable in the sense that it can be presented to any Moderator, not just the Moderator that issued it. Of course, this has implications for privacy, since the token necessarily reveals the identity of the Moderator that issued it. 
```

draft-authors-mole-vole-act.md
cjpatton: Cool, it sounds like this may actually be a useful feature. We'll mention this as a property of Ratatouille tokens in the intro :) Thanks Martin!

https://github.com/Moderation-of-unLinkable-Endorsements/internet-drafts/pull/47#discussion_r3776520187

draft-authors-mole-vole-act.md
cjpatton: I agree. There are also several reasons to generalize beyond a counter that can be "spent" and "refunded".:

1. See also @wbl's comment here: https://mailarchive.ietf.org/arch/msg/mole/e39YYAirqBvf6Ozq7FOEYL9BP5k/

2. @SamuelSchlesinger's also argued for generalization at a prior Privacy Pass interim:
https://datatracker.ietf.org/meeting/interim-2026-privacypass-01/materials/slides-interim-2026-privacypass-01-sessa-act-update-00

For now though I think it makes sense to keep Ratatouille in line with the language in ACT. Once make a change there, we can propagate it here. (By the way, I think @SamuelSchlesinger intends to move ACT into MoLE land but I'm not sure.)

draft-authors-mole-vole-act.md
cjpatton: Just to clarify: I read (b) to mean "check that the token's balance is greater than 0", which I think is required for ACTs no matter what. Perhaps this just needs to be clarified?

draft-authors-mole-vole-act.md
```diff
@@ -163,28 +260,844 @@ signature scheme and the VOLE-in-the-Head proof system (a component of the
...
+*unforgeability* (clients cannot mint or over-spend credits), however, rests on
+the strong Diffie-Hellman assumption that Shor's algorithm breaks. What
+remains, then, is to make unforgeability post-quantum as well.
+
+This document specifies Ratatouille, a **post-quantum** variant of ACT, whose
+security rests on assumptions believed to resist quantum attack: the hardness
+of structured multivariate quadratic (MQ) problems, via the Unbalanced Oil and
+Vinegar (UOV) signature scheme {{UOV}}, and the security of symmetric
+primitives (hash functions and pseudorandom generators), via the
+VOLE-in-the-Head (VOLEitH) zero-knowledge proof system {{FAEST}}. We build on
```
wbl, line 300: We're in a setting where the server could send a bunch of correlations, so designated verifier proofs could work. I'm not an expert, does this work out to anything faster?

draft-authors-mole-vole-act.md
```diff
@@ -193,7 +1106,89 @@ This document has no IANA actions.
 
 --- back
 
+# Performance {#performance}
```
cjpatton:
```suggestion
# Performance {#performance}
{:numbered="false"}
```

draft-authors-mole-vole-act.md
```diff
@@ -193,7 +1106,89 @@ This document has no IANA actions.
...
+in-circuit hash, and land in the same regime -- the residual gap is ISA and
+language (AVX C/C++ vs aarch64 Rust). Our RATA-MQ spend (prove 85 ms / verify 57
+ms) sits in the same regime once its widened-UOV map is folded with lift tables:
+the map is ~26x larger in coefficients, but the fold cost scales with the number
+of forms, not the map size, so the extra width costs little.
+
 # Acknowledgments
 {:numbered="false"}
-
-The use of Keccak-p[800,12] in this application was proposed by Bas Westerbaan.
```
cjpatton, line 199: Why remove this? @bwesterb definitely deserves credit / blame :)

draft-authors-mole-vole-act.md
cjpatton: Good question. Intuitively, the primitive we've built here is strictly more powerful than ACT: any Moderator can verify a Ratatouille token regardless of who issued it. Thus it seems reasonable that there should be a designated verifier VOLE approach that is more efficient.

We looked into this a bit and haven't turned up anything. I'd encourage you to noodle on it if you have time. Check out this paper: https://eprint.iacr.org/2021/076

A couple of observations:

1. Designed verifier VOLE requires the prover and verifier to interact in order to produce the correlations. Moreover the verifier has to keep some state between generating the VOLEs and verifying a proof.
2. We need something like the algebraic MAC used in ACT. Specifically, we need a way for the Moderator to produce a "blinded signature" that can be unblinded by the Client and revealed in the next spend. Ratatouille works differently: instead of producing of unblinding a signature, the Client proves knowledge of the signature. Producing this proof requires a signature signature instead of a MAC.

@SamuelSchlesinger had the same observation incidentally.

draft-authors-mole-vole-act.md
SamuelSchlesinger: Yeah, I mean it would be nice to have a DSL within which credits is one expressible state machine and compiles down to ~this efficient thing, but until we have that specializing to credits in the drafts makes sense to me. Not sure if I'll move ACT draft to this organization or keep it in my own GitHub, haven't decided yet. Once there is adoption of various drafts by a working group, they'll just live there anyways, so this shouldn't matter too much.

draft-authors-mole-vole-act.md
```diff
@@ -163,28 +260,844 @@ signature scheme and the VOLE-in-the-Head proof system (a component of the
...
+creating detailed logs of client behavior, raising significant privacy concerns
+in an era of increasing data protection awareness and regulation.
+
+The Anonymous Credit Token (ACT) protocol {{ACT}} resolves this tension by
+enabling credit-based systems without client tracking: an issuer can grant a
+client a token worth some number of credits, and the client can later spend
+those credits anonymously, in whole or in part, without the issuer being able
+to link the spend to the issuance or to any other spend by the same client.
+
+ACT {{ACT}} builds on BBS signatures over prime-order groups. Its
```
SamuelSchlesinger: BBS signatures is probably a bit of a misnomer here, it might be more accurate to say BBS MACs?

draft-authors-mole-vole-act.md
SamuelSchlesinger: I tried something like this, using VOLE instead of VOLEitH, in particular the ideal situation would be if a client could have a bunch of correlations "on hand" to do this. The gist is that there was no way to nicely do this while maintaining unlinkability and actually having a performance boost.

draft-authors-mole-vole-act.md
```diff
@@ -163,28 +260,844 @@ signature scheme and the VOLE-in-the-Head proof system (a component of the
...
+the UOV and MQ problems and to the security of standard symmetric primitives.
 
-# Conventions and Definitions
+This post-quantum security comes at a cost in **performance**: presentation
+proofs are larger than in the classical construction -- on the order of
+kilobytes to tens of kilobytes. Nonetheless, the protocol remains practical for
+modern web services, with efficient signing, verification, and proof
+generation.
+
+Unlike ACT, Ratatouille tokens are publicly verifiable in the sense that it can be presented to any Moderator, not just the Moderator that issued it. Of course, this has implications for privacy, since the token necessarily reveals the identity of the Moderator that issued it.
```
SamuelSchlesinger, line 349: I think it is important to note here that double-spend prevention only works if you use a single, shared database across redeemers.

draft-authors-mole-vole-act.md
```diff
@@ -193,7 +1106,89 @@ This document has no IANA actions.
...
+
+One-More-MAYO drops the symmetric primitive from the proven statement entirely,
+so it is both the smallest (6.7 KB) and fastest-to-show; the cost moves to
+`Sig1`. Its direct analogue is our One-More-UOV blind signature above (show 80
+ms / verify 47 ms / 4.4 KiB): both prove only a compact MQ preimage with no
+in-circuit hash, and land in the same regime -- the residual gap is ISA and
+language (AVX C/C++ vs aarch64 Rust). Our RATA-MQ spend (prove 85 ms / verify 57
+ms) sits in the same regime once its widened-UOV map is folded with lift tables:
+the map is ~26x larger in coefficients, but the fold cost scales with the number
+of forms, not the map size, so the extra width costs little.
```
SamuelSchlesinger: I'm not sure we need these benchmark results here. To me, it strikes me as unhelpful. In particular, it seems like we should list objective things like numbers of various operations, not talk about specific numbers or architectures we benchmarked on.

draft-authors-mole-vole-act.md
tnyuzg: Agree. Have changed this term.

draft-authors-mole-vole-act.md
tnyuzg: I think communication size is precise and universal across different computers.
I can try to let the code collect the number of different operations.

tnyuzg:
How should we unify the terminology, since the construction can be seen as a variant of ACT and a part of MoLE e.g. should we use the name "moderator"?

draft-authors-mole-vole-act.md
thibmeu: size is indeed helpful and generally true

draft-authors-mole-vole-act.md
```diff
@@ -163,28 +260,844 @@ signature scheme and the VOLE-in-the-Head proof system (a component of the
...
+  Input:
+    - dst: deployment domain-separation tag
+  Output:
+    - F: F_256^{n_com} -> F_256^k, G: F_256^{n_com} -> F_256^{m_uov-k}   // public quadratic maps
+  Steps:
+    1.  stream := SHAKE128("RATA-MQ-maps:" || dst)
+    2.  for each output equation i in [m_uov]:
+    3.      M_i := bytesIntoUpper(stream.next(n_com(n_com+1)/2))   // in F_256^{n_com * n_com}
+    4.  F := (M_0, ..., M_{k-1})                       // first k equations
+    5.  G := (M_k, ..., M_{m_uov-1})                   // remaining m_uov-k equations
```
SamuelSchlesinger: Lets make these general quadratic maps by adding linear terms as well. This fixes the forgery issue we discussed on Slack.

draft-authors-mole-vole-act.md
cjpatton: Left as a TODO

cjpatton (approved):
🦆

Issue #43, opened 2026-08-07
Token types?


cjpatton:
We currently think of IHAT and ACT as concrete protocols. It might be useful to reframe them as classes of protocols, of which there may be multiple instances. I suppose this is similar to how blind RSA and VOPRF are Privacy Pass token types, but I'm not 100% certain.

IHAT has a few "token types":

1. draft-authors-mole-crypto
2. Longfellow (as discussed in draft-jms-mole-protocol)
3. draft-authors-mole-vole-ihat

ACT has two:

1. draft-schlesinger-cfrg-act
2. draft-authors-mole-vole-act

What do folks think?

wbl:
I think this is a really good idea. Our (in progress) submission makes IHAT a definition to go along with IHAC (issuer hiding anonymous credentials). This means we need names...

wbl:
Rollatini?

thibmeu:
Registering tokens that mole needed within [privacy pass token registry](https://www.iana.org/assignments/privacy-pass/privacy-pass.xhtml) was discussed in the side meeting in Vienna. This is something that seems valuable overall to facilitate reuse of new token types within the privacy pass ecosystem.

The tradeoff is that we should not consider privacy pass to be the main use case. The interface fits well at the moment (both for endorsement and credential types). If that diverges too much from what privacy pass rfc defines (specifically RFC 9576 and RFC 9577), we should be ok to cut the tie down the line.

Registration should be relativaly straight forward by following the template provided in the [IANA consideration section of RFC 9578](https://datatracker.ietf.org/doc/html/rfc9578#name-privacy-pass-token-types)

cjpatton:
I don't think we necessarily need a **Privacy Pass** registry. If we decide to form a WG for MoLE, then we can create our own registry.

martinthomson:
Consider not doing the negotiation of type at the core protocol layer.  That is something that is better handled at a higher layer.

cjpatton:
@thibmeu pointed out to me today that, modulo reverse Privacy Pass, this notion of token types for IHAT and ACT corresponds exactly to "MoLE endorsements" and "MoLE credentials" in the protocols draft:
https://moderation-of-unlinkable-endorsements.github.io/internet-drafts/draft-jms-mole-protocols.html#section-9

What the protocols draft does today is already what I had in mind, except for potentially bikeshedding on naming, which I'm happy to save for later.

PR #48, opened 2026-08-14
vole-ihat: Complete the initial version

cjpatton:
Specify token issuance and redemption in enough detail to understand how the protocol works in broad strokes. Some details remain unspecified:

1. The wire format of messages sent in the protocol

2. The details of how the redemption proof is constructed. The circuit is described in a fair amount of detail, but it's not yet clear how the circuit is compiled into a proof. Also, an optimization that will be important in practice is to commit to intermediate states of the Keccak evaluation. We need to be precise about how this works.

Add estimates of the communication cost. An important observation (by Claude) is that the size of the redemption proof does not change that much as a function of the size of the Anchor Set. This is because most of the multiplications we're doing on the UOV maps depend on public constants rather than the witness, so these don't count towards the total proof size. We do have to commit to the selector vector, but this is tiny in comparison.

Update the constraints for the redemption proof. Claude came up with a set of "parity constraints" that appear to be a bit more robust. In particular the `b[i] * (b[i] - 1) = 0` checks don't actually prove that `b[i]` is `0` or `1` in a characteristic-2 field.

Minor things:

1. Zero-index vectors rather than start counting at 1. This is more consistent with Pythonic pseudocode, which we're approaching here.

2. Add a note about multishow (issue 45).

3. Define the interfaces for Keccak, UOV, and VOLEitH a bit more precisely than VOLE-ACT. We will eventually move this section to VOLE-ACT so that that document is the source of truth for all of our conventions.

4. Have Claude do a bit of copy editing.

### DIFF
--- a/draft-authors-mole-vole-ihat.md
+++ b/draft-authors-mole-vole-ihat.md
@@ -31,6 +31,7 @@ author:
 
 normative:
   IHAT: I-D.draft-authors-mole-crypto
+  VOLE-ACT: I-D.draft-authors-mole-vole-act
   FIPS202:  # TurboShortSHAKE128(M,d) = Sponge[Keccak-p[800, 12], pad10*1, 800–256](M||1111,d)
     title: "SHA-3 Standard: Permutation-Based Hash and Extendable-Output Functions"
     target: https://doi.org/10.6028/NIST.FIPS.202
@@ -148,6 +149,7 @@ normative:
 
 informative:
   ARCH: I-D.draft-jms-mole-architecture
+  SIGMA: I-D.draft-irtf-cfrg-sigma-protocols
   PoMFRIT:
     title: "Concretely Efficient Blind Signatures Based on VOLE-in-the-Head Proofs and the MAYO Trapdoor"
     target: https://eprint.iacr.org/2026/109
@@ -182,7 +184,7 @@ digital signature scheme and the VOLE-in-the-Head proof system (a component of
 the FAEST signature scheme). The anchor issues a token by signing a commitment
 to the client's state; to redeem a token, the client proves knowledge of the
 opening of this commitment and a signature produced by some anchor trusted by
-the moderator. The commitment is instantiated with Keccak-p[800,12].
+the moderator. The commitment is instantiated with `Keccak-p[800,12]`.
 
 --- middle
 
@@ -216,19 +218,18 @@ knowledge of a signature under some public key in the Anchor Set.
 
 Moussaka has significantly higher bandwidth cost compared to IHAT:
 
-1. each Anchor public key is `42.6` kilobytes (this corresponds to the uov-Ip
-   parameter set {{UOV}});
+1. each Anchor public key is 42.6KB (this corresponds to the uov-Ip parameter
+   set {{UOV}});
 
-1. `TODO` kilobytes are transmitted during issuance; and
+1. ~8KB are transmitted during issuance; and
 
-1. `TODO*L` kilobytes are transmitted during redemption,
+1. ~9KB are transmitted during redemption, plus a couple bytes per Anchor in
+   the Anchor Set.
 
-where `L` is the size of the Anchor Set. In particular, the size of the
-redemption proof scales linearily with the number of Anchors.
+> NOTE(cjpatton) These numbers aren't based on an actual implementation. They
+> are the result of an analysis from Claude Opus 5 looking at the FAEST spec
+> and an arithmetization of Keccak in a work in progress implementation.
 
-> NOTE(tnyuzg) We should be able to make redemption sublinear using the
-> OR proof of ia,cr/2025/113, Section 4.1.
->
 > NOTE(tnyuzg) We might be able to reduce the size of both VOLEitH proofs (one
 > for issuance and another for redemption) using the more invasive techniques
 > from {{PoMFRIT}}, Algorithm 2.
@@ -248,7 +249,25 @@ considerations for implementers and adopters.
 
 {::boilerplate bcp14-tagged}
 
-> TODO
+> TODO(cjpatton) Harmonize this section with {{VOLE-ACT}}.
+
+This document follows the same conventions and uses the same notation as
+{{VOLE-ACT}}.
+
+Vectors and sequences are indexed from zero: the elements of a length-`k` vector
+`v` are `v[0], ..., v[k-1]`. For an integer `k > 0`, we write `[k]` to denote
+the set of indices `{0, ..., k-1}`.
+
+We adopt the following notation for byte strings from {{IHAT}}. For two byte
+strings `x` and `y`, `x || y` denotes their concatenation. For a byte string
+`x`, `x[i..j]` denotes the substring of `x` that begins at its byte with index
+`i` and ends just before its byte with index `j`; its length is `j - i` bytes.
+
+We adopt the uov-Ip parameter set (NIST Level 1). That is, in the remainder of
+this document, let `n=112`, `m=44`, and `q=2^8`. Each element of `F_q` has a
+natural representation as a byte. We sometimes write `F_q^k` to denote the set
+of length-`k` byte strings, e.g., `r <- F_q^16` means to choose 16 random bytes
+and assign them to `r`.
 
 # Overview {#overview}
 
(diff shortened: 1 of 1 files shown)


draft-authors-mole-vole-ihat.md
thibmeu, line 18: note for easier reference down the line: could be good to rename the various crypto drafts with a pattern. something like `draft-<authors>-mole-endorsement-moussaka` so that one does not have to recall the mapping of acronyms to pet names

thibmeu, line 181: is this a variant of IHAT, or a new construction that satisfy endorsement?

draft-authors-mole-vole-ihat.md
```diff
@@ -281,70 +297,363 @@ The issuance protocol is as follows:
...
+   the `nf, r, ctx_red` from which `t` was computed, then sends `t, pf_iss` to
+   the Anchor.
 
 1. The Anchor verifies `pf_iss`, samples a solution `s` to `P(s) = t`, then
    sends `s` to the Client.
 
 1. The Client finalizes issuance by checking that `P(s) = t`.
 
-The process resembles blind signature issuance {{?RFC9474}}, The key difference
+The process resembles blind signature issuance {{?RFC9474}}. The key difference
```
thibmeu: nits

```suggestion
The process resembles rsa blind signature issuance {{?RFC9474}}. The key difference
```

draft-authors-mole-vole-ihat.md
```diff
@@ -281,70 +297,363 @@ The issuance protocol is as follows:
...
-for i in range(L):
-  for j in range(len(P[i])):
-     Q[j] += b[i] * P[i][j]
-~~~
-
-It then checks that `Q(s) == Com(nf, ctx_iss, ctx_red, r)` and that
-`b[i]*(b[i] - 1) == 0` for all `i in [L]`.
+in {{UOV}}. Namely, each public key is a homogeneous quadratic map, so that any
+linear combination of public keys is again a quadratic map; in particular, for
+any public key `P` it holds that `P(0^n) = 0^m`. The idea is to extend the
```
thibmeu: nits: this might be obvious, but having a definition that `b^n` is a vector of size `n` containing bytes set to `b` would be useful

draft-authors-mole-vole-ihat.md
```diff
@@ -281,70 +297,363 @@ The issuance protocol is as follows:
...
 
 # Protocol {#protocol}
 
-> TODO(cjpatton)
+> TODO(cjpatton) Specify wire formats for each of the messages. For now we will
+> ignore how things are encoded until we're more certain of the shape of the
+> protocol.
+
+The constant `VERSION`, an element of `F_q`, is used for versioning and is meant
+to be kept in sync with revisions to this document. Its current value SHALL be
```
thibmeu, line 399: is this a should or a must? i feel it's the later

```suggestion
to be kept in sync with revisions to this document. Its current value MUST be
```

draft-authors-mole-vole-ihat.md
```diff
@@ -281,70 +297,363 @@ The issuance protocol is as follows:
...
+    - ctx_iss: issuance context, in F_q^16
+    - ctx_red: redemption context, in F_q^16
+
+  Output:
+    - token: the issued token, or INVALID
+
+  Steps:
+    1. (nf, r) := pending_token
+    2. if P(s) != Com(nf, ctx_iss, ctx_red, r): return INVALID
+    3. token := (nf, r, s)
```
thibmeu, line 552: should the token solely be `(nf, r, s)` given that `ctx_iss and ctx_red` will be required to redeem it?

draft-authors-mole-vole-ihat.md
```diff
@@ -281,70 +297,363 @@ The issuance protocol is as follows:
...
+    5. request := (nf, pf_red)
+    6. return request
+~~~~
+
+### Finalization
+
+To complete the redemption request, the Moderator runs the following procedure.
+
+~~~~pseudocode
+FinalizeRedeem(P, request, ctx_iss, ctx_red):
```
thibmeu, line 650: nits: does it make sense to call this `verify`?

draft-authors-mole-vole-ihat.md
```diff
@@ -281,70 +297,363 @@ The issuance protocol is as follows:
...
+    - s: Anchor signature, in F_q^n
+    - ctx_iss: issuance context, in F_q^16
+    - ctx_red: redemption context, in F_q^16
+
+  Output:
+    - token: the issued token, or INVALID
+
+  Steps:
+    1. (nf, r) := pending_token
+    2. if P(s) != Com(nf, ctx_iss, ctx_red, r): return INVALID
```
thibmeu, line 551: maybe this could be extracted as a public verification method?

draft-authors-mole-vole-ihat.md
```diff
@@ -281,70 +297,363 @@ The issuance protocol is as follows:
...
+    - nf: nullifier
+    - ctx_iss: issuance context
+    - ctx_red: redemption context
+
+  Output:
+    - constraints: a sequence of constraints in the witness (r, s, b)
+
+  Steps:
+    1. constraints := []
+    2. Q := Select(b, P)
```
thibmeu: `b` should be in the input set

draft-authors-mole-vole-ihat.md
thibmeu, line 297: given we are doing an IHAT comparison, worth noting that `ctx_iss` and `ctx_red` differ in sizes. based on the later definition

```
    - nf: the nullifier, in F_q^24
    - ctx_iss: issuance context, in F_q^16
    - ctx_red: redemption context, in F_q^16
```

draft-authors-mole-vole-ihat.md
```diff
@@ -281,70 +297,363 @@ The issuance protocol is as follows:
...
+    - request: redemption request
+
+  Steps:
+    1. (nf, r, s) := token
+    2. b := 0^L  // selector vector
+    3. b[i] := 1
+    4. pf_red := VOLEitH.Prove(
+         (r, s, b),                          // witness
+         (P, nf, ctx_iss, ctx_red),          // public
+         RedeemRel(P, nf, tx_iss, ctx_red),  // relation
```
thibmeu:
```suggestion
         RedeemRel(P, nf, ctx_iss, ctx_red),  // relation
```

thibmeu (approved):
overall looks good. i think this could even stand on its own without refering to IHAT, simply stating it's a PQ unforgeable version of endorsement (or issuer-hiding if we want to decorrelate from mole)

draft-authors-mole-vole-ihat.md
tnyuzg: I don't need the property `P(0^n) = 0^m` now

cjpatton:
> overall looks good. i think this could even stand on its own without refering to IHAT, simply stating it's a PQ unforgeable version of endorsement (or issuer-hiding if we want to decorrelate from mole)

Makes sense. I'll stick with the current phrasing for now, as it's consistent with Ratatouille. Eventually we might resolve this in a different way for both: see #43.

draft-authors-mole-vole-ihat.md
cjpatton: The current name is a nod to @SamuelSchlesinger original proposal for ACT: VOLE-ACT. It also anticipates resolving #43 by thinking of Moussaka as an instantiation of the IHAT token type. (Likewise for ACT.)

cjpatton: We might resolve #43 by defining IHAT to be a "token type", of which Moussaka will be one instance. Right now I'm just calling it a variant/alternative to IHAT for consistency.

cjpatton: Good catch. This is already discussed below, but I'll add a forward reference to that section here.

draft-authors-mole-vole-ihat.md
cjpatton: Good idea, done.

draft-authors-mole-vole-ihat.md
```diff
@@ -281,70 +297,363 @@ The issuance protocol is as follows:
...
+redemption proof is the opening `r`, the signature `s`, and a selector vector
+`b[0], ..., b[L-1]`, where `b[u] = 1` if `u = i` and `0` otherwise. The circuit
+first selects the public key as `Q := Select(b, P)`, where `Select()` is as
+defined in {{overview}}. The first constraint is that `Q(s) = Com(nf, ctx_iss,
+ctx_red, r)`.
+
+Next, the circuit needs to prove that the selector vector is well-formed, that
+is, exactly one entry is set. Because `F_q` has characteristic `2`, the weight
+constraint `sum(b) = 1` only tells us that the number of nonzero entries is
+odd. We also need some parity constraints, namely `sum(b[0..u+1]) * b[u+1] = 0`
```
tnyuzg: what about using
sum(b) = 1;
for all i in [L]: (b[i]-1) b[i] = 0
to constrain the selector vector b?
there should be less addition gate

draft-authors-mole-vole-ihat.md
cjpatton: Not obvious, it needs to be defined. This will be defined in {{VOLE-ACT}}.

cjpatton: Good catch, removed.

draft-authors-mole-vole-ihat.md
tnyuzg: We can let the anchor index i be in the input, compute b = [0,0,..1,...,0] inside the function

draft-authors-mole-vole-ihat.md
cjpatton: These mean the same thing: https://datatracker.ietf.org/doc/html/rfc2119#section-1

draft-authors-mole-vole-ihat.md
cjpatton: Why? It's not an API that would be needed by any party.

draft-authors-mole-vole-ihat.md
cjpatton: The alternative, which I've implemented below, is to pass the context strings to the functions used for redemption.

Either is fine I think. I think the current approach encourages a cleaner implementation: the caller of `RequestRedeem()` should know in advance the redemption and issuance context the Moderator expects.

draft-authors-mole-vole-ihat.md
cjpatton: Done.

draft-authors-mole-vole-ihat.md
```diff
@@ -281,70 +297,363 @@ The issuance protocol is as follows:
...
+    3. constraints.append("Q(s) = Com(nf, ctx_iss, ctx_red, r)")
+    6. for u in [L-1]:
+    7.   constraints.append("sum(b[0..u+1]) * b[u+1] = 0")
+    8. constraints.append("sum(b) = 1")
+    9. return constraints
+~~~~
+
+> TODO An optimization that will be important for practice is to commit
+> intermediate states of the Keccak evaluation to the witness. We need to
+> define precisely how this works.
```
tnyuzg: for now we additionally commit to internal states S_4 and S_8, which are of gap-4
The reason we didn't use gap-6 e.g. S_6 and S_12 with the meet-in-the-middle trick is that
we need to commit to the last round i.e. S_12 to make the trick work
So compared to gap-4 S_0, S_4, S_8 (which we don't need to commit to last round), S_0, S_6, S_12 don't save much

draft-authors-mole-vole-ihat.md
cjpatton: Maybe! I chose "finalize" because the output is a nullifier rather than a boolean, and it has a nice symmetry with "finalize issue" above. I don't have a strong preference here.

draft-authors-mole-vole-ihat.md
cjpatton: Welllll, it's more accurate to say that `b` is part of the witness than `i`, since it's literally the value we commit to. In fact, `b` is formally a variable here: For the prover it refers to the witness; and for the verifier,it indexes into the proof. We need to think about how to finesse this.

Actually, I'm now realizing that the sigma-protocols draft probably sets a precdence for this already. I'll take a look at that and try to align our relations with theirs.

draft-authors-mole-vole-ihat.md
cjpatton: As discussed 2026/8/14: the "party constraints" avoid having to emulate "big num" subtraction in the circuit. It instead uses addition that's native to a characteristic-2 field.

draft-authors-mole-vole-ihat.md
cjpatton: Ok, I pushed a change that aligns us a bit closer to the sigma protocols draft. I think this is good enough for now.

tnyuzg (approved):

SamuelSchlesinger (approved):

draft-authors-mole-vole-ihat.md
```diff
@@ -216,19 +218,18 @@ knowledge of a signature under some public key in the Anchor Set.
...
 
-1. `TODO*L` kilobytes are transmitted during redemption,
+1. ~9KB are transmitted during redemption, plus a couple bytes per Anchor in
+   the Anchor Set.
 
-where `L` is the size of the Anchor Set. In particular, the size of the
-redemption proof scales linearily with the number of Anchors.
+> NOTE(cjpatton) These numbers aren't based on an actual implementation. They
+> are the result of an analysis from Claude Opus 5 looking at the FAEST spec
+> and an arithmetization of Keccak in an work in progress implementation.
```
cjpatton:
```suggestion
> and an arithmetization of Keccak in a work in progress implementation.
```

draft-authors-mole-vole-ihat.md
```diff
@@ -267,84 +286,394 @@ A UOV public key is a map `P` from `F_q^n` to `F_q^m`. A UOV signature is a
...
+We write `s := UOV.SPre(td, t)` to denote sampling a pre-image of target `t`
+under `P`, i.e., an `s` for which `P(s) = t`. This is the same procedure as used
+in the signing algorithm in {{UOV}}, Figure 2.
+
+We write `pf := VOLEitH.Prove(R, X, W)` to denote proving knowledge of a
+witness `W` for which the pair `(X, W)` is in the relation `R`, where `X` is
+the instance, using the VOLE-in-the-Head proof system defined in {{FAEST}}. We
+write `v := VOLEitH.Verify(R, X, pf)` to denote verification of `pf` under the
+same relation and instance, where the output `v` is a bit. The components of
+`X` and `W` are passed in the order in which they are declared by `R`.
```
cjpatton:
```suggestion
same relation and instance, where the output `v` is a bit.
```

draft-authors-mole-vole-ihat.md
```diff
@@ -267,84 +286,394 @@ A UOV public key is a map `P` from `F_q^n` to `F_q^m`. A UOV signature is a
...
+
+Relations are specified in the style of {{Section 3.4 of SIGMA}} as a sequence
+of constraints. Each constraint is an equation of the instance and witness
+variables. Unlike {{SIGMA}}, these equations need not be linear: each side of
+the equation is a polynomial of arbitrary. However, because the degree of the
```
cjpatton:
```suggestion
the equation is a polynomial of arbitrary degree. However, because the degree of the
```

draft-authors-mole-vole-ihat.md
```diff
@@ -267,84 +286,394 @@ A UOV public key is a map `P` from `F_q^n` to `F_q^m`. A UOV signature is a
...
+polynomial influences the size of the proof, the relation needs to be designed
+to balance the size of the witness with the degree of the constraints.
+
+> TODO(cjpatton) Figure out how to cite {{FAEST}} more precisely. Ideally it
+> specifies a compiler of constraints into a proof, as in {{SIGMA}}, but this
+> is currently beyond the scope of what the FAEST spec does. The degree of
+> constraints is coupled tightly to the proof generation procedure, which means
+> I think we'll end up pulling some details of the proof construction into this
+> document,
```
cjpatton:
```suggestion
> document.
```

draft-authors-mole-vole-ihat.md
```diff
@@ -267,84 +286,394 @@ A UOV public key is a map `P` from `F_q^n` to `F_q^m`. A UOV signature is a
...
+    - ctx_iss: issuance context, in F_q^16
+
+  Output:
+    - s: Anchor signature, in F_q^n, or INVALID
+
+  Steps:
+    1. (t, pf_iss) := request  // commitment, issuance proof
+    2. v := VOLEitH.Verify(
+         IssueRelation,  // relation
+         (ctx_iss, t),   // instance
```
cjpatton:
```suggestion
         (t, ctx_iss),   // instance
```

PR #51, opened 2026-08-17
vole-act: Pass of overview

cjpatton:
Based on #50 (merge that first).

Harmonize the overview with the scope and tone of Moussaka. Also, add a NOTE about removing one of the hashes in the spend proof and add some notes about security.


thibmeu (approved):
overall unblocking so that we can have a clear verison to review

PR #57, opened 2026-08-19
vole-act: Opinionated changes

cjpatton:
Harmonize domain separation between Ratatouille and Moussaka, removing the "application context" from the former. Also, add issuance context to Ratatouille. Its length is 16 bytes to match Moussaka.

Minor things:

1. Hardcode `q=256` as in Moussaka. We're unlikely to make a different choice (such as `q=16`, which matches the uov-Is parameter set) given how nice it is to work with bytes.

2. Replace Greek letter names with more meaningful names, e.g., `nu` -> `nf_len`.

3. Add variable to define the length of context strings.

4.  Update TODOs.


draft-authors-mole-vole-act.md
```diff
@@ -523,8 +522,13 @@ by a concrete configuration, and {{instantiations}} specifies two concrete
 configurations, called Ratatouille-KP800 and Ratatouille-MQ. A deployment MUST
 fix its configuration for the lifetime of an Issuer key.
 
-> TODO(cjpatton) Bind the variant to key derivation and the various key
-> operations to mitigate confusion attacks properly.
+A global constant `VERSION in F_q` is defined. Its value SHALL be `0`. This
```
SamuelSchlesinger: nit: `VERSION` in `F_q`.

draft-authors-mole-vole-act.md
```diff
@@ -523,8 +522,13 @@ by a concrete configuration, and {{instantiations}} specifies two concrete
...
+constant is used for domain separation and is meant to be kept in sync with
+revisions to this document. The overall template for domain separation is as
+follows: the first four bytes of each permutation input identify the protocol;
+the next byte is the document version (`VERSION`); the next byte indicates the
+usage (commit, tag, or MQ map derivation); and the next byte indicates the
+configuration (KP800 or MQ). The domain is therefore seven bytes in total.
```
SamuelSchlesinger: This strikes me as too small for domain separation, should we not permit a deployment ID which is a longer string?

draft-authors-mole-vole-act.md
cjpatton: I took the liberty of including it in the ticks because we also use `in` as a keyword in our pseudocde :) 

Happy to fix though!

draft-authors-mole-vole-act.md
cjpatton: Unfortunately we don't have a ton of bytes to work with. In any case, I think a deployment ID could be covered by the issuance context?

draft-authors-mole-vole-act.md
SamuelSchlesinger: Reasonable, leave as is.

cjpatton: Done, here and everywhere else.

draft-authors-mole-vole-act.md
SamuelSchlesinger: I believe I misunderstood, I thought we were talking about domain separation at the protocol level.

cjpatton: I'm thinking more about modeling `KP800()` as a random oracle. When working in the ROM, it's usually helpful to be able to easily distinguish RO queries made for different purposes.

I'm also thinking about cross protocol attacks. Imagine an Anchor re-used its Moussaka key for Ratatouille. We don't want to encourage this obviously, but if it happens, we want to make sure a message sent in one protocol can't be used to attack the other.

I'm not sure the domain separation scheme I'm sufficient for either of these purposes ... we would need to work out the security proof to be certain. I'm just making an educated guess based on prior experience.

SamuelSchlesinger (approved):

PR #60, opened 2026-08-21
crypto: Add Stacking Sigmas and Endorsement Redemption

deep-inder:
Specified endorsement redemption in the crypto draft: additive key rerandomization plus an issuer-hiding proof over the Moderator's Anchor Set, built from Stacking Sigmas (eprint 2021/422).

Specified: key rerandomization; the Schnorr branch proof; the partially binding vector commitment from discrete log composed into a binary tree; the Fiat-Shamir transcript; the algorithms ProveIssuer, VerifyIssuer, Redeem and VerifyRedemption, and the Redemption wire format; and the security considerations as prose. 

Not yet specified, and marked TODO in the text: formal security statements and the corresponding reductions, and test vectors.

### DIFF
--- a/draft-authors-mole-crypto.md
+++ b/draft-authors-mole-crypto.md
@@ -24,13 +24,14 @@ venue:
 
 author:
  -
-    fullname: John Doe
-    organization: ACME
-    email: john@example.com
+    fullname: Deep Inder Mohan
+    organization: Georgia Institute of Technology
+    email: dmohan@gatech.edu
 
 normative:
   ARCH: I-D.draft-jms-mole-architecture
   PROTOCOLS: I-D.draft-jms-mole-protocols
+  HTTP-TRANSPORT: I-D.draft-jms-mole-http-transport
   HASH2CURVE: RFC9380
   I2OSP: RFC8017
   OPRF: RFC9497
@@ -55,6 +56,64 @@ normative:
 
 informative:
   SIGMA: I-D.draft-irtf-cfrg-sigma-protocols
+  CDS94:
+    title: "Proofs of Partial Knowledge and Simplified Design of Witness Hiding Protocols"
+    target: https://doi.org/10.1007/3-540-48658-5_19
+    date: 1994
+    seriesinfo:
+      "CRYPTO": "1994"
+    author:
+      -
+        ins: R. Cramer
+        name: Ronald Cramer
+      -
+        ins: I. Damgard
+        name: Ivan Damgard
+      -
+        ins: B. Schoenmakers
+        name: Berry Schoenmakers
+  FFKLLS26:
+    title: "Issuer-Hiding BBS-Based Anonymous Credentials without Policy Keys"
+    target: https://eprint.iacr.org/2026/870
+    date: 2026
+    author:
+      -
+        ins: A. Flamini
+        name: Andrea Flamini
+      -
+        ins: F. Friedrichs
+        name: Felix Friedrichs
+      -
+        ins: J. Katz
+        name: Jonathan Katz
+      -
+        ins: W. Ladd
+        name: Watson Ladd
+      -
+        ins: A. Lehmann
+        name: Anja Lehmann
+      -
+        ins: C. Sefranek
+        name: Cavit Sefranek
+  STACKSIG:
+    title: "Stacking Sigmas: A Framework to Compose Sigma-Protocols for Disjunctions"
+    target: https://eprint.iacr.org/2021/422
+    date: 2022
+    seriesinfo:
+      "EUROCRYPT": "2022"
+    author:
+      -
+        ins: A. Goel
+        name: Aarushi Goel
+      -
+        ins: M. Green
+        name: Matthew Green
+      -
+        ins: M. Hall-Andersen
+        name: Mathias Hall-Andersen
+      -
+        ins: G. Kaptchuk
+        name: Gabriel Kaptchuk
   TESSZHU:
     title: "Short Pairing-Free Blind Signatures with Exponential Security"
     target: https://eprint.iacr.org/2022/047
@@ -120,12 +179,14 @@ This document is a work in progress. This revision specifies:
 * the endorsement issuance protocol, that is, the four algorithms `Commit`,
   `Challenge`, `Respond`, and `Finalize`, along with the wire messages they
   exchange, and the endorsement verification equation ({{issuance}});
+* endorsement redemption, that is, key rerandomization, the issuer-hiding
+  proof over a Moderator's Anchor Set, whose size is logarithmic in that of the
+  Anchor Set, and the algorithms `Redeem` and `VerifyRedemption`
+  ({{redemption}});
 * two ciphersuites, over P-256 and ristretto255 ({{ciphersuites}}).
 
 The following are **not yet specified** and are marked as such in the text:
 
-* endorsement redemption, including the issuer-hiding proof over a
-  Moderator's Anchor Set ({{redemption}});
 * the full security considerations ({{security-considerations}});
 * test vectors ({{test-vectors}}).
 
@@ -134,7 +195,11 @@ The following are **not yet specified** and are marked as such in the text:
 > Client sends the first message. In the construction specified here the Anchor
 > sends the first message, so the algorithms are named `Commit`, `Challenge`,
 > `Respond`, and `Finalize`. **TODO:** rename these in {{PROTOCOLS}}. The number
-> of HTTP exchanges (two) and the endorsement type are unchanged.
+> of HTTP exchanges (two) and the endorsement type are unchanged. For the same
+> reason the redemption algorithms are named `Redeem` and `VerifyRedemption`
+> rather than `Present` and `Verify`: this document already uses `Verify` for
+> endorsement verification under a known key ({{verify}}), and {{ARCH}} calls
+> the operation a redemption.
 
 > **Editorial note.** This document binds an Endorsement to two contexts, an
 > *issuance context* and a *redemption context* ({{context-binding}}), whereas
@@ -153,7 +218,10 @@ are used as defined in {{ARCH}}.
 
 Unless otherwise specified, this document encodes protocol messages in TLS
 notation ({{Section 3 of TLS13}}). Moreover, all constants are in network byte
-order.
+order. This document also uses the variable-size vector convention `<V>`
+defined in {{HTTP-TRANSPORT}}: the length prefix of such a vector is a
+variable-length integer in its minimum-size encoding, so its width depends on
+the length of the contents it carries.
 
 The following functions and notation are used throughout this document.
 
@@ -165,6 +233,10 @@ For a byte string `x`, `x[i..j]` denotes the substring of `x` that begins at
 its byte with index `i` and ends just before its byte with index `j`, where
 indices start at zero. Its length is `j - i` bytes.
 
+For a list `x`, `x[i]` denotes its element at index `i`, counting from zero,
+`x[i..j]` denotes the sublist from index `i` up to but not including index `j`,
+and `len(x)` denotes the number of elements it holds.
+
 `I2OSP(x, xLen)` converts a nonnegative integer `x` into a byte string of
 length `xLen` in big-endian byte order, as described in {{I2OSP}}.
 
@@ -172,6 +244,10 @@ length `xLen` in big-endian byte order, as described in {{I2OSP}}.
 them with a cryptographically secure random number generator. It is the only
 source of randomness in this document: every other value that has to be
 unpredictable is derived from its output ({{derive-scalar}}).
+
+`Seed(x, k)` denotes the `k`-th seed in a byte string of concatenated seeds,
+that is `x[k * Nseed .. (k + 1) * Nseed]`, with `k` counted from zero.
+{{ciphersuites}} fixes the seed length `Nseed`.
 
 String values in monospace and quotes, such as `"Challenge"`, are ASCII string
 literals and do not include a terminating NUL byte.
@@ -608,9 +684,9 @@ def Commit(skA, ctx_iss):
   Z = CreateContextBase(ctx_iss)
 
   rand = random(3 * Nseed)
-  a = DeriveScalar(rand[0 .. Nseed], "a")
-  t = DeriveScalar(rand[Nseed .. 2 * Nseed], "t")
-  y = DeriveScalar(rand[2 * Nseed .. 3 * Nseed], "y")
+  a = DeriveScalar(Seed(rand, 0), "a")
+  t = DeriveScalar(Seed(rand, 1), "t")
+  y = DeriveScalar(Seed(rand, 2), "y")
 
   A = G.ScalarMultGen(a)
   C = G.ScalarMultGen(t) + y * Z
@@ -670,10 +746,10 @@ def Challenge(pkA, ctx_iss, ctx_red, commitment):
   nf = rand[0 .. Nn]
   seeds = rand[Nn .. Nn + 4 * Nseed]
 
-  r1     = DeriveScalar(seeds[0 .. Nseed], "r1")
-  r2     = DeriveScalar(seeds[Nseed .. 2 * Nseed], "r2")
-  gamma1 = DeriveScalar(seeds[2 * Nseed .. 3 * Nseed], "gamma1")
-  gamma2 = DeriveScalar(seeds[3 * Nseed .. 4 * Nseed], "gamma2")
+  r1     = DeriveScalar(Seed(seeds, 0), "r1")
+  r2     = DeriveScalar(Seed(seeds, 1), "r2")
+  gamma1 = DeriveScalar(Seed(seeds, 2), "gamma1")
+  gamma2 = DeriveScalar(Seed(seeds, 3), "gamma2")
 
   m = Message(nf, ctx_red)
   gamma = gamma1 * G.ScalarInverse(gamma2)
@@ -717,7 +793,7 @@ def ComputeChallenge(ctx_iss, commitment, m):
   return c
 ~~~
 
-Two challenge values appear here and it matters which is which. `c` is computed
+Two challenge values appear here: `c` is computed
 over the *blinded* commitment and is the value that ends up in the Endorsement
 ({{finalize}}); it never leaves the Client. The `challenge` message the Anchor
 receives is its blinded form `c * gamma2`, and the Anchor cannot recover `c`
@@ -843,12 +919,10 @@ The Endorsement is held privately by the Client until it is redeemed
 ## Endorsement Verification {#verify}
 
 An Endorsement is publicly verifiable under the issuing Anchor's public key.
-Verification requires no interaction with the Anchor and no pairing.
 
-The two contexts are inputs to `Verify`, not fields of the Endorsement. A
+The two contexts are inputs to `Verify` in addition to the Endorsement. A
 verifier therefore states the pair it is willing to accept and learns whether
-the Endorsement was issued under it, rather than being told by the Client which
-pair to check against.
+the Endorsement was issued under it.
 
 Input:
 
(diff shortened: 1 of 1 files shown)


draft-authors-mole-crypto.md
```diff
@@ -1036,29 +1102,770 @@ invalid Endorsement.
...
+hidden from the verifier, is binding, while every other position can afterwards
+be opened to any value. The Client commits the branch commitment of `index` at
+the binding position, and after the challenge is known it opens every other
+position to the branch commitment that the shared response determines. The
+proof of one Schnorr branch is therefore honest; which one is hidden by the
+commitment key.
+
+This document uses the commitment of Section 5.2 of {{STACKSIG}} over pairs,
+composed into a binary tree as described in Section 5.3 of {{STACKSIG}}, so
+that a commitment to `2^q` values costs one commitment key and one opening per
```
wbl, line 1273: We should make a TODO to handle non-powers of two. I know this is ugly, and nonblocking

draft-authors-mole-crypto.md
```diff
@@ -1036,29 +1102,770 @@ invalid Endorsement.
...
+`G0` is a fixed parameter of the ciphersuite, since the domain separation tag of
+`HashToGroup` carries `ctx_proto` ({{config}}). It is derived by hashing a
+constant so that no party knows its discrete logarithm; see
+{{security-considerations}}.
+
+A node of the tree commits two values with a commitment key `ck`, which is a
+single group element. The two values are committed under two derived bases:
+
+~~~
+def CreateNodeBases(ck):
```
wbl, line 1293: Oh, is this how the construction shakes out for the 1-of-2 scheme when you apply the start of the paper? We used the permutation scheme in the appendix.

draft-authors-mole-crypto.md
```diff
@@ -1036,29 +1102,770 @@ invalid Endorsement.
...
+  for i in range(N):
+    T = BranchCommitment(proof_challenge, response, Y[i])
+    below[i] = CompressValue(0, G.SerializeElement(T))
+
+  # below holds level j - 1 and above level j, of N / 2^j values.
+  for j in range(1, q + 1):
+    opening = (openings[2 * (j - 1)], openings[2 * (j - 1) + 1])
+    above = []
+    for k in range(N / 2^j):
+      node = CommitNode(commitment_keys[j - 1], opening,
```
wbl, line 1642: Should we note the shared work at each level?

draft-authors-mole-crypto.md
```diff
@@ -1036,29 +1102,770 @@ invalid Endorsement.
...
+} Redemption;
+~~~
+
+`shown_endorsement` uses the `Endorsement` structure of
+{{endorsement-encoding}}, with `s` carrying `s_hat`. It is a valid Endorsement
+under `rerandomized_key`, which is the point of {{rerandomization}}; it is not
+the Endorsement the Client stored, and the Client MUST NOT send that one.
+
+`commitment_keys` MUST hold exactly `q` elements and `openings` exactly
+`2 * q` scalars, where `q` is the depth of the tree determined by the Anchor Set
```
wbl, line 1852: The permutation based one uses only q scalars

draft-authors-mole-crypto.md
```diff
@@ -1224,12 +2031,116 @@ Derived blinding factors:
...
+  (Section 5.2 of {{STACKSIG}}): a Client that can open one position of a node
+  two ways knows the discrete logarithm of that position's base, and a Client
+  that can do so at both positions knows that of `G0`, since the two bases sum
+  to `G0 + G0`. `G0` MUST therefore be the `HashToGroup` output specified in
+  {{pbvc}} and MUST NOT be chosen, negotiated, or supplied by any party: a
+  Client that knew its discrete logarithm could equivocate every position of
+  every node, open the binding leaf to whatever the challenge required, and
+  produce accepting redemptions holding no Endorsement at all. The same
+  reduction is why a node commits its two values under two *separate* Pedersen
+  commitments rather than one: with a single commitment over both bases, a
```
wbl, line 2101: That's not exactly the reason: rather we'd have an equation that makes life hard for showing it binds

draft-authors-mole-crypto.md
```diff
@@ -1224,12 +2031,116 @@ Derived blinding factors:
...
+: The proof is logarithmic in the size of the Anchor Set: two scalars, plus one
+  element and two scalars for each of the `q` levels of the tree
+  ({{redemption-wire}}). Computation is not. Both the prover and the verifier
+  evaluate every branch and every node, which is `N` branch commitments and
+  `N - 1` node commitments, so each performs a number of scalar multiplications
+  linear in `N`. A large Anchor Set is therefore cheap in bandwidth and not in
+  CPU, which reverses the tradeoff of the linear disjunction {{CDS94}} for
+  bandwidth but not for work; {{FFKLLS26}} notes the same for its own
+  instantiations. Two consequences for deployments: the linear disjunction is
+  smaller for Anchor Sets of five keys or fewer, and because the tree is padded
```
wbl, line 2128: I have crossover before 4 with the implementation choices I made

wbl (approved):
Looks good. I think there are some choices to think about/useful enhancements
- Ristretto and a permutation based approach to getting the keys can shrink the openings, not sure how it apply to P256 if we want to stick with that
- Non powers of two are handled with padding, but there are more efficient ways that make only n-1 compressions for n elements.
- The security discussion ignores OMUF as far as I can tell

draft-authors-mole-crypto.md
deep-inder: Yes, but that approach requires some group permutation that I'm not sure I quite understand, and I wasn't fully sure how to write it out in the spec. Happy to switch to that approach, but will need some help in figuring out how to write it.

draft-authors-mole-crypto.md
```diff
@@ -992,16 +1063,11 @@ Client cannot guess, and so collide with, another Client's session. An Anchor
 that receives a `ChallengeMessage` whose `session_id` does not correspond to one
 of its open sessions MUST raise a `SessionError`.
 
-The session identifier MUST NOT be bound into the challenge transcript, and it is
-not an input to any algorithm in {{issuance}}. It is a value the Anchor chose and
+The session identifier MUST NOT be bound into the challenge transcript, and it MUST NOT be an input to any algorithm in {{issuance}}. It is a value the Anchor chose and
```
cjpatton, line 1066: nit: wrap at 80 columns

draft-authors-mole-crypto.md
```diff
@@ -1036,29 +1102,770 @@ invalid Endorsement.
...
->   Moderator's, and is unrelated to the issuance challenge of {{challenge}}.
-> * *What the Moderator learns.* Both contexts are inputs to verification
->   ({{verify}}), so a Moderator states the pair it accepts rather than being
->   told it by the Client; a Client whose Endorsement was issued under any other
->   pair simply fails. The redemption reveals `nf`, which the Moderator checks
->   it has not seen before.
+A Client redeems an Endorsement at a Moderator. This document calls the
+operation a redemption, following {{ARCH}}, which reserves the term
+Presentation for the Credential flow; the message a Client sends is nonetheless
+carried in the `Presentation` structure of {{PROTOCOLS}} ({{redemption-wire}}).
```
cjpatton, line 1108: I don't think this needs to be said.

```suggestion
A Client redeems an Endorsement at a Moderator.
```

draft-authors-mole-crypto.md
```diff
@@ -1036,29 +1102,770 @@ invalid Endorsement.
...
+   index = the position of the Client's Anchor in anchor_set
+   redemption = Redeem(anchor_set, index, endorsement,
+                       ctx_iss, ctx_red, challenge_digest)
+
+                             redemption
+                              -------->
+
+              verified = VerifyRedemption(anchor_set, redemption,
+                                          ctx_iss, ctx_red,
+                                          challenge_digest)
```
cjpatton, line 1139: syntax: For Moussaka we we went with `IssueRequest()`/`IssueResponse()`/`FinalizeIssue()` for issuance and `RedeemRequest()`/`FinalizeRedeem()` redemption. The extra move for issuance means we can't align there, but I think it would be worth aligning on redemption.

I think my only strongly held view here is that the Moderator's redemption function should output the nullifier it's meant to use for double spend prevention (or else raise an error if redemption fails).

draft-authors-mole-crypto.md
```diff
@@ -1036,29 +1102,770 @@ invalid Endorsement.
...
+~~~
+{: #fig-redemption title="Endorsement redemption overview"}
+
+`anchor_set` is the list of Anchor public keys the Moderator accepts.
+The Moderator carries it in the `Challenge` structure of {{PROTOCOLS}}, and its
+**order is significant**: proof branches are matched to keys by position. Both
```
cjpatton, line 1145: Redundant 
```suggestion
The Moderator carries it in the `Challenge` structure of {{PROTOCOLS}}. Both
```

draft-authors-mole-crypto.md
```diff
@@ -1036,29 +1102,770 @@ invalid Endorsement.
...
+parties MUST use the same list in the same order, and a Moderator MUST reject a
+redemption whose proof is not shaped for exactly the Anchor Set it offered
+({{redemption-wire}}).
```
cjpatton, line 1148: "proof is shaped" is confusing. In any case, isn't it the case that the proof wouldn't verify if the anchor sets don't match? In this case, MUST is inappropriate.
```suggestion
parties MUST use the same list in the same order.
```

draft-authors-mole-crypto.md
```diff
@@ -1036,29 +1102,770 @@ invalid Endorsement.
...
+`index` is the position in `anchor_set` of the public key of the Anchor that
+issued the Endorsement. A Client whose Anchor does not appear in `anchor_set`
+cannot redeem at this Moderator and MUST NOT try: the Endorsement will not
+verify under any key it can prove membership for.
+
+## Key Rerandomization {#rerandomization}
+
+The Client shifts the Anchor's public key by a secret scalar `delta` of its own
+choosing and adapts the signature to the shifted key. Writing `pkA` for
+`anchor_set[index]` and `(c, s, y, t, nf)` for the Endorsement:
```
cjpatton, line 1165: nit: In Moussaka we put the nullifier in the first position of the token tuple.

draft-authors-mole-crypto.md
```diff
@@ -1036,29 +1102,770 @@ invalid Endorsement.
...
+~~~
+
+Every other value `Verify` recomputes is untouched, so a Moderator can check
+the signature by running `Verify` ({{verify}}) with `X_hat` in place of `pkA`.
+
+This shift is *additive*, so issuance is left
+completely unmodified and the unforgeability of {{issuance}} carries over
+({{security-considerations}}); and it is *public*, in the sense that `X_hat` on
+its own says nothing about which key it was derived from, since `delta * B` is
+uniformly distributed over the group. On its own it also proves nothing: any
```
cjpatton, line 1189: "On its own it also proves nothing": This sounds like a Claudism to me

draft-authors-mole-crypto.md
```diff
@@ -1036,29 +1102,770 @@ invalid Endorsement.
...
+Client can produce a uniform `X_hat`. This is why the client must additionally provide an issuer-hiding proof.
```
cjpatton, line 1190: Here you want to answer the question: "proof of what"?

Also, wrap at 80 columns

draft-authors-mole-crypto.md
```diff
@@ -1036,29 +1102,770 @@ invalid Endorsement.
...
+forces one branch to have been answered honestly. {{FFKLLS26}} applies the same
+two compositions, for the same purpose, to a pairing-based credential.
+
+Both compositions are used with the same rerandomization ({{rerandomization}}),
+and switching between them changes neither issuance nor {{verify}}.
+
+Each branch is the discrete logarithm proof of {{SIGMA}}, and the composition
+below could be expressed in that framework. It is written out here instead, so
+that this document fixes the transcript and the encodings without depending on
+work in progress. **TODO:** revisit once {{SIGMA}} is stable.
```
cjpatton, line 1216:
```suggestion
work in progress. 

> TODO Revisit once {{SIGMA}} is stable.
```

draft-authors-mole-crypto.md
```diff
@@ -1036,29 +1102,770 @@ invalid Endorsement.
...
+
+Throughout this section and its subsections, `x / y` denotes integer division
+of nonnegative integers, that is the quotient rounded down, and `x mod y` the
+remainder.
```
cjpatton, line 1220: Worth hoisting to conventions?

draft-authors-mole-crypto.md
```diff
@@ -1036,29 +1102,770 @@ invalid Endorsement.
...
+~~~
+
+of which the Client can answer branch `index`, where `Y[index] = delta * B`.
+
+A single branch is proved by the usual three moves: the Client draws a nonce
+`r` and commits `T = r * B`; the verifier sends a challenge `c`; the Client
+answers `z = r - c * delta`; and the verifier checks that
+`T = c * Y[index] + z * B`.
+
+Read the other way round, that check *determines* the commitment from the
```
cjpatton, line 1237: nit: too much emphasis?

```suggestion
Read the other way around, this check determines the commitment from the
```

draft-authors-mole-crypto.md
```diff
@@ -24,13 +24,14 @@ venue:
 
 author:
  -
```
cjpatton, line 26: This document needs more editors. @wbl and @SamuelSchlesinger correct?

draft-authors-mole-crypto.md
```diff
@@ -1036,29 +1102,770 @@ invalid Endorsement.
...
+~~~
+
+Two properties of this branch proof are what allow the composition below, and
+{{STACKSIG}} calls a sigma protocol with both of them *stackable*. First, a
+verifying commitment can be computed for *any* statement from a challenge and a
+response, by the function above, without knowing a witness; this is the
+extended honest-verifier zero-knowledge property. Second, the response of an
+honest run is a uniformly distributed scalar whatever the statement and the
+witness, so one response can be reused across branches without revealing which
+branch produced it.
```
cjpatton, line 1252: I don't follow. "a uniformly distributed scalar whatever the statement and the witness"

draft-authors-mole-crypto.md
```diff
@@ -1036,29 +1102,770 @@ invalid Endorsement.
...
+
+### Partially Binding Commitments {#pbvc}
+
+A *partially binding vector commitment* {{STACKSIG}} commits to a vector of
+values such that one position, chosen when the commitment key is generated and
+hidden from the verifier, is binding, while every other position can afterwards
+be opened to any value. The Client commits the branch commitment of `index` at
+the binding position, and after the challenge is known it opens every other
+position to the branch commitment that the shared response determines. The
+proof of one Schnorr branch is therefore honest; which one is hidden by the
```
cjpatton, line 1268: what do you mean by honest here?

draft-authors-mole-crypto.md
```diff
@@ -1036,29 +1102,770 @@ invalid Endorsement.
...
+
+  return G.HashToScalar(proof_transcript)
+~~~
+
+`n` is prefixed and `Element` encodings are fixed-length, so
+`anchor_set_enc` and `ck_enc` are unambiguous without length prefixes of their
+own; `q`, and with it
+the number of commitment keys, is determined by `n`. The label `"IssuerProof"`
+separates this transcript from the issuance transcript of {{challenge}}, which
+is hashed with the same function.
```
cjpatton, line 1450: Wrap at 80 columns

```suggestion
`anchor_set_enc` and `ck_enc` are unambiguous without length prefixes of their
own; `q`, and with it the number of commitment keys, is determined by `n`. The
label `"IssuerProof"` separates this transcript from the issuance transcript of
{{challenge}}, which is hashed with the same function.
```

draft-authors-mole-crypto.md
```diff
@@ -1036,29 +1102,770 @@ invalid Endorsement.
...
+
+Unlike `ComputeChallenge`, this function has no nonzero requirement: a challenge
```
cjpatton, line 1452: What does "nonzero requirement" mean?

draft-authors-mole-crypto.md
cjpatton: I don't understand, does this mean the length of the anchor set needs to be a power of 2? Could we just pad with fake keys up to the next power of 2?

draft-authors-mole-crypto.md
```diff
@@ -1036,29 +1102,770 @@ invalid Endorsement.
...
+`VerifyIssuer` returns `false` rather than raising an error, so that it is a
+total predicate on its inputs, as `Verify` ({{verify}}) is. Two of its three
+rejections are defensive restatements of its input types: the lengths of
+`commitment_keys` and `openings` are fixed by the Anchor Set, and a redemption
+whose vectors have any other length is rejected before this algorithm is
+reached, when it is deserialized ({{redemption-wire}}). The third, `n < 2`, is
+not a property of the redemption at all but of the Moderator's own Anchor Set;
+reaching it means the Moderator is misconfigured ({{verify-redemption}}).
+
+An honest proof verifies. On branch `index`,
```
cjpatton, line 1669: Verifies what? The equation below? The "." is throwing me I think

draft-authors-mole-crypto.md
```diff
@@ -1224,12 +2031,116 @@ Derived blinding factors:
...
+  branch's statement. Nothing in `Redeem` gives a Client a signature it did not
+  already hold: `X_hat` and `s_hat` are computed from values it has, and a
+  Client that could produce an accepted redemption without an Endorsement from
+  some Anchor in `anchor_set` would yield a forgery. **TODO:** this reduction is
+  stated, not written out.
+
+Issuer hiding:
+: A redemption is distributed identically no matter which Anchor in `anchor_set`
+  issued the Endorsement, so a Moderator, an Anchor, and the two colluding
+  learn only that some key in `anchor_set` was used. Three facts combine.
```
cjpatton, line 2059: Combine to make what?

cjpatton (commented):
Thanks @deep-inder! I mostly didn't follow the math, but left some comments mostly about presentation. On that I have two high level bits of feedback:

1. Stacked Sigmas is quite involved. I think the specification could use some sign posting to make it easier to digest. First, we should include a high level overview somewhere. Second, every once in a while we need to remind the reader of how the part they're reading relate to the bigger picture.

2. Not for this PR, but: I'm quite worried about the maintenance burden we're creating for ourselves. In the likely scenario that we need to make changes, it's going to be very difficult to propagate those changes correctly throughout the entire document. In the near future, I think we should transcribe the pseudocode here into a pure Python implementation that we can write tests for and generate test vectors from. We can then replace the psuedcode in the draft with Python snippets from the reference implementation. [This worked out really well for VDAF, which has a similar level of complexity](https://www.ietf.org/archive/id/draft-irtf-cfrg-vdaf-22.html#section-7.3.3). For what it's worth, this is something that Claude or GPT would be able to handle quite easily.

draft-authors-mole-crypto.md
wbl: We do pad now (it's later in the doc), but that does a whole bunch of extra work. It's possible to change the shape of the tree to do exactly n-1 reductions for n inputs, with the same number of openings (logarithmic), but the text to do that is a bit complex. It's worth doing as otherwise you end up with really uneven performance as you cross power of two boundaries. I was lazy in the implementation now.

draft-authors-mole-crypto.md
```diff
@@ -1036,29 +1102,770 @@ invalid Endorsement.
...
+
+  proof_challenge, response, commitment_keys, openings =
+    ProveIssuer(anchor_set, index, delta, X_hat, shown, ctx_iss,
+                ctx_red, challenge_digest, rand[Nseed .. nrand])
+
+  return Redemption(X_hat, shown, proof_challenge, response,
+                    commitment_keys, openings)
+~~~
+
+A Client MUST NOT redeem against an Anchor Set of fewer than two keys, and
```
cjpatton, line 1747: It would be useful if this were at least possible mechanically. We're likely going to have an early test deployment or two for which there is just one anchor in the world.

thibmeu:
Created https://github.com/deep-inder/internet-drafts/pull/1 to address editorial and API feedback. @deep-inder : feel free to review/merge. another path is to merge the present PR, and then add my commit on top

thibmeu:
merged in #64

PR #58, opened 2026-08-19
protocols: Warn about Anchor Set intersection attacks

cjpatton:

PR #63, opened 2026-09-02
protocols: Warn about Anchor Set intersection attacks

thibmeu:
Adapted from #58 for the current configuration model, given Chris is out this week.

### DIFF
--- a/draft-jms-mole-protocols.md
+++ b/draft-jms-mole-protocols.md
@@ -786,6 +786,17 @@ TODO. The list to cover:
    fingerprinting vector (with {{ARCHITECTURE}}).
 3. Epoch width versus anonymity set size.
 
+During redemption of an Endorsement, the Client uses the accepted Anchor Set
+from the Moderator's authenticated configuration. If the Client does not have
+an Endorsement issued by one of the Anchors in the Anchor Set, it must either
+abort the redemption flow or pause it until it can obtain a suitable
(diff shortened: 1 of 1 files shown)


draft-jms-mole-protocols.md
```diff
@@ -786,6 +786,17 @@ TODO. The list to cover:
...
+During redemption of an Endorsement, the Client uses the accepted Anchor Set
+from the Moderator's authenticated configuration. If the Client does not have
+an Endorsement issued by one of the Anchors in the Anchor Set, it must either
+abort the redemption flow or pause it until it can obtain a suitable
+Endorsement. The Client must take care to ensure its actions in this case do not
+inadvertently reveal the issuing Anchor of an accepted Endorsement. For example,
+if the Client initiates two concurrent redemption flows, the Moderator can
+select Anchor Sets that differ by just one Anchor. If one flow aborts but the
+other does not, then the Moderator immediately learns the Anchor of the accepted
+Endorsement.
```
SamuelSchlesinger, line 798: I mean, isn't that why the moderator must commit to an Anchor set? It feels like we should mention the mitigation as well with a SHOULD.

thibmeu: this is a privacy consideration which is to be drafted still. I've added your suggestion as a TODO

cjpatton (approved):

PR #64, opened 2026-09-03
Crypto draft

thibmeu:
Essentially #60 , but avoiding having deep toi merge https://github.com/deep-inder/internet-drafts/pull/1 into their own branch before this becomes mergeable

@wbl @deep-inder


wbl (approved):
There's issues, but probably best to merge and then interate on sections.

thibmeu:
the floor is yours now @wbl

PR #65, opened 2026-09-03
crypto: switch to compact form of vector commitments, handle 1

wbl:
Still work to do: this doesn't handle randomness in the same way because I couldn't figure it out, but I say merge and we can hack away. There may also be references to things that are made wrong by shrinking things, and we'll have to handle zero length vectors in the encoding.

### DIFF
--- a/draft-authors-mole-crypto.md
+++ b/draft-authors-mole-crypto.md
@@ -1241,6 +1241,17 @@ challenge and the response:
 ~~~
 def BranchCommitment(proof_challenge, response, Y):
   return proof_challenge * Y + G.ScalarMultGen(response)
+
+def Statements(anchor_set, X_hat):
+n = len(anchor_set)
+  q = 0
+  while 2^q < n:
+    q = q + 1
+
+  for i in range(len(anchor_set):
+    Y[i] = X_hat - anchor_set[i]
+
+return (Y, q)
 ~~~
 
 Two properties of this branch proof are what allow the composition below, and
@@ -1259,6 +1270,8 @@ construction. Nothing is proved by the responses themselves. What has to be
 enforced is that the Client fixed the commitment of one branch *before* the
 challenge existed, and could not afterwards change it.
 
+
+
 ### Partially Binding Commitments {#pbvc}
 
 A *partially binding vector commitment* {{STACKSIG}} commits to a vector of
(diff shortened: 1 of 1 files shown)


draft-authors-mole-crypto.md
```diff
@@ -1271,141 +1271,58 @@ branch at the binding position uses the actual witness and a commitment fixed
...
-and does so by choosing which:
+# the return value is (Q, secret) where
+# secret can be used to equivocate the non
+# binding direction
 
-~~~
-def CreateCommitmentKey(trapdoor, equivocal):
-  E = G.ScalarMultGen(trapdoor)
+def GenerateStep(bind_direction):
+    secret = G.RandomScalar()
```
thibmeu, line 1310: nits: should this be seeded?

draft-authors-mole-crypto.md
```diff
@@ -1937,6 +1800,32 @@ DeserializeScalar(buf):
...
+  left = buf[0:17] # 17 bytes
+  right = buf[17:33] # 16 bytes
+
+  left ^= SHAKE256(right, 17)
+  right ^ = SHAKE256(left, 16)
+  left ^= SHAKE256(right, 17)
+  right ^ = SHAKE256(left, 16)
+
+  buf = left || right
+  buf[0] &= 0x01
```
thibmeu: any concerns with masking buf?

draft-authors-mole-crypto.md
```diff
@@ -1271,141 +1271,58 @@ branch at the binding position uses the actual witness and a commitment fixed
...
-
-  if equivocal == 0:
-    return (r_left - trapdoor * v, r_right)
-  else:
-    return (r_left, r_right - trapdoor * v)
+def VecCommit(V, Qi, rands):
+    if len(V) == 1 :
+        return V[0]
+    for i:=0; i<len(V)/2; i++
+        V'[i] = CommitStep(Q[0], V[2*i], V[2*i+1], rands[0])
```
thibmeu:
```suggestion
        V'[i] = CommitStep(Qi[0], V[2*i], V[2*i+1], rands[0])
```

draft-authors-mole-crypto.md
```diff
@@ -1271,141 +1271,58 @@ branch at the binding position uses the actual witness and a commitment fixed
...
+    if len(V) % 2 == 1 :
+        V'[len(V)/2+1] = V[len(V)-1]
+    return VecCOmmit(V', Qi[1:], rands[1:])
```
thibmeu:
```suggestion
    return VecCommit(V', Qi[1:], rands[1:])
```

draft-authors-mole-crypto.md
```diff
@@ -1494,29 +1411,13 @@ Errors: `DeriveError`
 def ProveIssuer(anchor_set, index, delta, X_hat, endorsement, ctx_iss,
                 ctx_red, challenge_digest, rand):
   (Y, q) = Statements(anchor_set, X_hat)
```
thibmeu, line 1426: you are deleting `Statements` as part of this commit. is this line still correct?

thibmeu (approved):
the brevity is nice. I think you'll need to run `make fix-lint` or approve the various whitespace deletion suggestion to make CI happy

my comments are mostly about ensuring the algorithm read correctly. I would not be able to analyse their cryptographic correctness

draft-authors-mole-crypto.md
wbl: It should be but I need to figure out how we carry all the seeds needed through every step

draft-authors-mole-crypto.md
wbl: Yes, this needs to happen before the call to SHAKE256 gets put in with right. The reason we do this is we want a permutation on 257 bits, not 33 bytes. Will fix and push before merge or right after.

draft-authors-mole-crypto.md
wbl: I'll need to undo that deletion

PR #66, opened 2026-09-09
Connect IHAT redemption to the protocol and HTTP drafts

SamuelSchlesinger:
Map IHAT redemption to the existing cryptographic API and make the HTTP Moderator challenge carry the common RedemptionChallenge. Remove stale interface and context notes, and correct the draft URL and TLS reference.

Prepping for adding ACTs.


cjpatton (approved):
Nice

PR #67, opened 2026-09-09
put ACT in its own draft

SamuelSchlesinger:
this puts ACT in its own draft and lines it up with the IHAT Python interfaces. shared definitions now point to IHAT, and the model and key helpers live in `poc/` too.

stacked on #70. the later branches fill in the proofs and the rest of the demo.

One small point: I am using the word "signed" here instead of "authenticated". I just think it is more intuitive for readers, and when we shift to VOLEitH it will be an actual signature.


dennisjackson (approved):

draft-authors-mole-act.md
```diff
@@ -0,0 +1,763 @@
...
+presentations to each other or to issuance. Anonymous Credit Tokens (ACT)
+provide a numeric balance as this state ({{credential-scheme}}).
+
+The Moderator is both the issuer and the verifier. A Client obtains an initial
+Credential, spends credits, and finalizes the refund that replaces the spent
+Credential. {{PROTOCOLS}} maps these operations to the MoLE credential APIs,
+specifies when issuance is authorized, and supplies the credential and spend
+contexts. This document treats those contexts as opaque byte strings.
+
+# Conventions and Definitions
```
cjpatton, line 150: Consider deduplicating any conventions or preliminaries that this document has in common with IHAT. In other words, just refer to {{IHAT}} for everything that's shared, sort of like we do for {{ARCH}}:

> The terms Client, Moderator, and Credential are used as defined in
> {{ARCH}}.

This might force you to transcribe the pseudocode to Python, but that may be worth doing anyway :)))

draft-authors-mole-act.md
```diff
@@ -0,0 +1,763 @@
...
+
+The Moderator publishes `SerializeElement(pkM)` in its configuration; see
+{{PROTOCOLS}}.
+
+# The Credential Scheme {#credential-scheme}
+
+A MoLE Credential is an Anonymous Credit Token (ACT). It is a
+keyed-verification anonymous credential {{KVAC}} over a privately verifiable,
+pairing-free BBS-style signature {{BBS}} {{TZ23}}, whose hidden state is a
+*balance*: an integer number of credits. A Moderator issues a Credential with
```
cjpatton: ?

```suggestion
*balance*: a positive integer number of credits. A Moderator issues a Credential with
```

draft-authors-mole-act.md
```diff
@@ -0,0 +1,763 @@
...
+an initial balance, and the Client later *spends* from it. A spend reveals a
+public amount `s` and proves, in zero knowledge, that the Credential holds at
+least `s` credits and has not been spent before. In the same exchange the
+Moderator issues a *refund*: a fresh Credential whose balance is the remainder
+plus a Moderator-chosen return amount `t`. A spend may also declare a public
+*top-up* `a`, in which case the refund may return up to `s + a`, raising the
```
cjpatton: It might be useful to align notation with {{VOLE-ACT}} here. I.e., I think we would say `x` instead of `a`?

cjpatton (approved):
Very nice, a few thoughts in comments but this is great.

poc/src/act/protocol.py
```diff
@@ -0,0 +1,56 @@
+"""ACT algorithms transcribed from draft-authors-mole-act.md.
+
+Configuration is fixed for a run. The demo uses an 8-bit balance; the
```
cjpatton: This comment about the demo is sort of irrelevant here. Perhaps remove from the doc string?

poc/src/act/py.typed
cjpatton, line 1: I wonder if we can deduplicate this and `src/ihat/py.typed` by moving `py.typed` up to `src`?

poc/src/ihat/derivation.py
```diff
@@ -0,0 +1,60 @@
...
+    DeriveScalar: Callable[[bytes, bytes], Scalar]
+    DeriveKeyPair: Callable[[bytes, bytes], tuple[Scalar, Element]]
+    GenerateKeyPair: Callable[[], tuple[Scalar, Element]]
+
+
+def BindDerivation(
+    G: P256Group,
+    ctx_proto: bytes,
+    Nseed: int,
+    random: Callable[[int], bytes],
```
cjpatton: is this always going to be `common.random` Could we just import it here?

poc/tests/test_draft_snippets.py
cjpatton, line 1: nice one :)

poc/src/ihat/derivation.py
cjpatton, line 1: It took me a bit, but I think I see what this module is accomplishing: Both ACT and IHAT have some code that depends on a global constant, so this module deduplicates the code that's shared.

This is reasonable, but makes me think trying to treat these as global variables is complicating the code structure unnecessarily.

We could move DeriveScalar and DeriveKeyPair and GenerateKeyPair to methods on the group. This changes the notation slightly from the draft: we woiuld write "G.DeriveScalar" instead of "DeriveScalar" and so on and pass is ctx_proto as a function parameter. But this seems worth it if we can simplify this module a bit.

The random() function is always going to be the same across IHAT and ACT, so I don't see why we need to include it as input to BindDerivation.

The Seed() function is a bit more complicated, since `Nseed` seems to differ between IHAT and ACT. We could:

1. Make `Nseed` the same and move it and `Seed` to the common module. This would mean revisiting whether we're cool with pseudorandom values for IHAT. I'm pretty sure we are (cc @wbl).
2. Just redefine `Seed()` in both IHAT and ACT. This is probably the lowest friction, even if it means decduplicating a bit of code.

poc/src/act/protocol.py
SamuelSchlesinger: Done

poc/src/ihat/derivation.py
SamuelSchlesinger: Yep, done

poc/src/ihat/derivation.py
SamuelSchlesinger: Agreed, moved the derivation methods to G and removed the binding module. Also shared Seed and Nseed because both schemes use 48 bytes.

draft-authors-mole-act.md
SamuelSchlesinger: Yes, lets align once we can see both in full. VOLE's `x` is the amount returned, which is `t` here. `a` is the top-up allowance. Clarified the text so hopefully that is clearer.

draft-authors-mole-act.md
SamuelSchlesinger: Done, nonnegative because zero is valid.

draft-authors-mole-act.md
SamuelSchlesinger: Yes, that's underway as you can see

poc/src/ihat/ciphersuite.py
```diff
@@ -195,6 +199,28 @@ def HashToScalar(self, value: bytes, *, DST: bytes | None = None) -> P256Scalar:
         uniform = expand_message_xmd(value, dst, 48)
         return P256Scalar(int.from_bytes(uniform, "big") % ORDER)
 
+    def DeriveScalar(self, seed: bytes, info: bytes) -> P256Scalar:
+        if len(seed) != Nseed:
```
wbl: We end up using this for arbitrary length strings/group elements some places. I also feel we should rely on info for distinguishing internally derived scalars over other calls in other places, but this PR doesn't need to do that: this whole comment is more to get discussion of what we want to see.

draft-authors-mole-ihat.md
jkatz2, line 96: K.

jkatz2, line 97: Karla, not Felix

jkatz2, line 109: Marek, not Cavit

draft-authors-mole-act.md
```diff
@@ -0,0 +1,522 @@
...
+        ins: J. Traore
+        name: Jacques Traore
+
+...
+
+--- abstract
+
+This document specifies Anonymous Credit Tokens (ACT), a keyed-verification
+anonymous credential scheme with a hidden credit balance. A Client spends a
+public amount and receives a fresh Credential for the remaining balance and
```
jkatz2, line 127: nit: why is Credential capitalized here and in general?

draft-authors-mole-act.md
```diff
@@ -0,0 +1,522 @@
...
+This document defines its cryptographic operations, encodings, and security
+requirements. The MoLE protocols use ACT as a Credential scheme.
+
+--- middle
+
+# Introduction
+
+MoLE Credentials carry per-Client state that a Moderator tests and updates at
+each presentation without learning it, and without being able to link
+presentations to each other or to issuance. Anonymous Credit Tokens (ACT)
```
jkatz2, line 141: (ACTs)

SamuelSchlesinger:
Looks like I had the wrong target here.

PR #68, opened 2026-09-11
ihat: Convert pseudocode to Python

cjpatton:
Add an (incomplete) reference implementation of IHAT and replace the pseudocode in the draft with snippets from the reference implementation.


poc/src/ihat/ciphersuite.py
```diff
@@ -0,0 +1,285 @@
...
+from .common import I2OSP
+
+
+FIELD_MODULUS = 0xFFFFFFFF00000001000000000000000000000000FFFFFFFFFFFFFFFFFFFFFFFF
+ORDER = 0xFFFFFFFF00000000FFFFFFFFFFFFFFFFBCE6FAADA7179E84F3B9CAC2FC632551
+CURVE_A = FIELD_MODULUS - 3
+CURVE_B = 0x5AC635D8AA3A93E7B3EBBD55769886BC651D06B0CC53B0F63BCE3C3E27D2604B
+SSWU_Z = FIELD_MODULUS - 10
+GENERATOR_X = 0x6B17D1F2E12C4247F8BCE6E563A440F277037D812DEB33A0F4A13945D898C296
+GENERATOR_Y = 0x4FE342E2FE1A7F9B8EE7EB4A7C0F9E162BCE33576B315ECECBB6406837BF51F5
```
cjpatton, line 21: Note: This is for scalar field arithmetic. EccPoint doesn't export this directly as far as I know.

poc/src/ihat/ciphersuite.py
```diff
@@ -0,0 +1,285 @@
...
+        raise ValueError("invalid expand_message_xmd output length")
+
+    dst_prime = dst + I2OSP(len(dst), 1)
+    b0 = _sha256(bytes(64) + msg + I2OSP(length, 2) + b"\x00" + dst_prime)
+    blocks = [_sha256(b0 + b"\x01" + dst_prime)]
+    for counter in range(2, ell + 1):
+        blocks.append(
+            _sha256(_xor(b0, blocks[-1]) + I2OSP(counter, 1) + dst_prime)
+        )
+    return b"".join(blocks)[:length]
```
cjpatton, line 112: The hash-to-curve draft has an implementation, but we can't use it directly:
https://github.com/cfrg/draft-irtf-cfrg-hash-to-curve/tree/main/poc

We confirm interop with test vectors.

poc/src/ihat/ciphersuite.py
```diff
@@ -0,0 +1,285 @@
...
+    ) -> Element[P256SHA256]:
+        dst = DST if DST is not None else b"HashToGroup-" + self.ctx_proto
+        uniform = expand_message_xmd(value, dst, 96)
+        u0 = int.from_bytes(uniform[:48], "big") % FIELD_MODULUS
+        u1 = int.from_bytes(uniform[48:], "big") % FIELD_MODULUS
+        x0, y0 = self._map_to_curve_simple_swu(u0)
+        x1, y1 = self._map_to_curve_simple_swu(u1)
+        q0: Element[P256SHA256] = Element(EccPoint(x0, y0, curve="p256"))
+        q1: Element[P256SHA256] = Element(EccPoint(x1, y1, curve="p256"))
+        return q0 + q1
```
cjpatton, line 191: This is also covered by test vectors.

poc/src/ihat/ciphersuite.py
```diff
@@ -0,0 +1,284 @@
...
+    @staticmethod
+    def _sqrt_ratio(u: int, v: int) -> tuple[bool, int]:
+        p = FIELD_MODULUS
+        tv1 = v * v % p
+        tv2 = u * v % p
+        tv1 = tv1 * tv2 % p
+        y1 = pow(tv1, (p - 3) // 4, p) * tv2 % p
+        y2 = y1 * pow(10, (p + 1) // 4, p) % p
+        is_square = y1 * y1 % p * v % p == u
+        return is_square, y1 if is_square else y2
```
cjpatton, line 284: Likewise, these are reimplemented from the hash-to-curve draft. Test vectors confirm interop.

That said, I'd appreciate extra eyes to make sure the implementation is correct.

draft-authors-mole-crypto.md
```diff
@@ -630,63 +567,31 @@ algorithm takes that message as a single input. The messages are the
...
-  return state, commitment
+    return (AnchorState(a, y, t), Commitment(A, C))
 ~~~
 
 The Anchor stores `state` for the duration of the session and sends `commitment`
-to the Client in a `CommitMessage` ({{wire}}). Note that `skA` is not used in
-`Commit`; it appears in the interface because an implementation MAY choose to
-carry it in the session state rather than reload it in `Respond`.
+to the Client in a `CommitMessage` ({{wire}}). The signing key is not needed
+until `Respond`.
```
cjpatton, line 595: I don't see much point in hinting at a potential design decision for an implementation. Besides, the keyword MAY is inappropriate here.

draft-authors-mole-crypto.md
```diff
@@ -1373,63 +1253,51 @@ privileged by it.
...
+
+    proof_challenge = ComputeProofChallenge(
+        anchor_set,
+        X_hat,
+        endorsement,
+        ctx_iss,
+        ctx_red,
+        challenge_digest,
+        commitment_keys,
+        root,
```
cjpatton, line 1290: @wbl This was previously `EncodeNode(root)`, but `EncodeNode` was undefined. Perhaps `U16Prefixed(root)` is what you want?

wbl: The root of the tree will always be a group element in this design (if n=1 it's a commitment), so I don't know why we need to length prefix it? Maybe I'm missing something. EncodeNode definitely was not what I meant.

poc/src/ihat/ciphersuite.py
wbl: Should I then change P to reduce number of distinct hash functions needed?

poc/src/ihat/ciphersuite.py
```diff
@@ -0,0 +1,284 @@
...
+        if len(buf) != self.Ns:
+            raise DeserializeError("invalid P-256 scalar length")
+        scalar = int.from_bytes(buf, "big")
+        if scalar >= ORDER:
+            raise DeserializeError("P-256 scalar is out of range")
+        return P256Scalar(scalar)
+
+    def P(self, element: Element[P256SHA256]) -> Element[P256SHA256]:
+        # SPEC: The P-256 permutation description has an incorrect SEC1 prefix
+        # range and does not precisely define its cycle-walking inverse.
```
wbl, line 241: What I meant to say was the SEC1 prefix of 0x02, 0x03 gets mapped to 0x0, 0x01, and then mapped back for decoding. Probably needs a tweak to make that clearer. The inverse is going to be apply the inverse on the strings until you get to a group element: again, some text might help which I can add (but will be away for a bit to do so).

draft-authors-mole-crypto.md
cjpatton: Ah Ok, this may actually have been from @deep-inder. If you don't think we need it, then we don't :)

poc/src/ihat/protocol.py
```diff
@@ -0,0 +1,559 @@
...
+
+    Z = CreateContextBase(ctx_iss)
+    m = Message(nf, ctx_red)
+
+    C = G.ScalarMultGen(t) + y * Z
+    A = G.ScalarMultGen(s) - (c * y) * pkA
+    commitment = Commitment(A, C)
+
+    return c == ComputeChallenge(ctx_iss, commitment, m)
+
```
wbl, line 255: This is where I would expect to see token randomization

poc/src/ihat/protocol.py
```diff
@@ -0,0 +1,559 @@
...
+def CommitStep(
+    Q: Element,
+    left: bytes,
+    right: bytes,
+    randomness: Scalar,
+) -> bytes:
+    C = (
+        randomness * B
+        + DeriveScalar(left, b"left") * Q
```
wbl, line 287: This ends up making left and right fixed length because of the check in DeriveScalar, and they don't have to be.

poc/src/ihat/protocol.py
```diff
@@ -0,0 +1,559 @@
...
+        return V[0]
+
+    V_prime = []
+    for i in range(len(V) // 2):
+        V_prime.append(CommitStep(Qi[0], V[2 * i], V[2 * i + 1], rands[0]))
+
+    if len(V) % 2 == 1:
+        # NOTE(GPT 5.6 Sol) The draft uses index len(V) / 2 + 1, which leaves
+        # an undefined entry in a zero-based vector. Carrying the final value
+        # contiguously is the apparent intent. The binding path through such a
```
wbl, line 325: Sol should be smarter / drawing a few diagrams should clear it up, but the code is right.

poc/src/ihat/protocol.py
```diff
@@ -0,0 +1,559 @@
...
+        + ck_enc
+        + root
+        + b"IssuerProof"
+    )
+
+    return G.HashToScalar(proof_transcript)
+
+
+def GenerateVecBind(q, index, rand):
+    # TODO(wbl) Define the types of the inputs and outputs and specify this
```
wbl, line 374: This is supposed to figure out what path to take, and take it in a vector of length q to bind at index index.

poc/src/ihat/protocol.py
```diff
@@ -0,0 +1,559 @@
...
+    # function. It is currently referenced by ProveIssuer.
+    raise NotImplementedError("GenerateVecBind is not specified")
+
+
+def CommitValAtPlace(q, index, value):
```
wbl, line 379: Will fill in: the naive form can just pad out the rest of the vector with whatever, but there is an optimized one

poc/src/ihat/protocol.py
```diff
@@ -0,0 +1,559 @@
...
+    root: bytes,
+) -> Scalar:
+    (c, s_hat, y, t, nf) = endorsement
+    n = len(anchor_set)
+
+    anchor_set_enc = b""
+    for i in range(n):
+        anchor_set_enc += G.SerializeElement(anchor_set[i])
+
+    ck_enc = b""
```
wbl, line 349: I feel very confused now. Isn't this the same as the anchor set, or derivable from it in some way?

poc/src/ihat/protocol.py
```diff
@@ -0,0 +1,559 @@
...
+    endorsement: Endorsement,
+    ctx_iss: bytes,
+    ctx_red: bytes,
+    challenge_digest: bytes,
+    rand: bytes,
+) -> tuple[Scalar, Scalar, Sequence[Element], Sequence[Scalar]]:
+    (Y, q) = Statements(anchor_set, X_hat)
+    if not 0 <= index < len(anchor_set):
+        raise ValueError("index is outside the Anchor Set")
+    if len(rand) != (3 * q + 1) * Nseed:
```
wbl, line 405: We'll probably need less. The need to propagate the randomness consumption through the calls is annoying, maybe we can have a better interface?

wbl (approved):
I haven't checked the SWU implementation in detail, and there's some things highlighted I'll need to address (or someone can hop in on), but this is a good step.

poc/src/ihat/ciphersuite.py
cjpatton: I think that would be a good idea :)

cjpatton: https://github.com/Moderation-of-unLinkable-Endorsements/internet-drafts/issues/71

poc/src/ihat/ciphersuite.py
cjpatton: will leave as a TODO here for now.

poc/src/ihat/protocol.py
cjpatton: Actually `Verify()` is transcribed from the draft without any changes. This is a standalone function that's used in the draft to clarify how the proof works, it's not actually used by the Moderator.

poc/src/ihat/protocol.py
cjpatton: I'll delete the comment.

poc/src/ihat/protocol.py
cjpatton: This function was previously as follows in the draft:

```
def CommitStep(Q, left, right, randomness):
    C = randomness*G.Generator()+G.DeriveScalar(left)*Q+G.DeriveScalar(right)*P(Q)
    return G.SerializeElement(C)
```

`G.DeriveScalar()` was not defined, but the intent appears to be `DeriveScalar()`. In other words I don't think I've actually changed anything here.

poc/src/ihat/protocol.py
cjpatton: All I did was transcribe verbatim. Here's the original:

https://github.com/Moderation-of-unLinkable-Endorsements/internet-drafts/blame/99c650f11da28a8e5b8718c450e9a672ef38b198/draft-authors-mole-crypto.md#L1334-L1361

poc/src/ihat/protocol.py
cjpatton: Got it! It wasn't specified in the draft so I just left as a stub here.

poc/src/ihat/protocol.py
cjpatton: If naive is simpler to implement but still asymptotically logarithmic, then I say keep it simple

poc/src/ihat/protocol.py
cjpatton: Agreed. Again, I'm just copying verbatim from the spec.

poc/src/ihat/protocol.py
wbl: We might have to tweak derive scalar then, I'll do it in another pass

poc/src/ihat/protocol.py
wbl: I'll make a note to look at this harder

PR #69, opened 2026-09-11
ihat: Editorial pass

cjpatton:
Editorial changes. I've also dropped some comments with some non-editorial things I wanted to discuss.


draft-authors-mole-crypto.md
```diff
@@ -465,13 +473,17 @@ several scalars from one seed instead would cap their joint entropy at the
 length of that seed, which the unlinkability argument of
 {{security-considerations}} does not permit. See {{randomness}}.
```
cjpatton, line 468: This seems more strict than necessary. For instance, couldn't scalars derived from the same `seed` but `info`s be treated as independent, pseudorandom numbers? Why do they need to be statistically close to random?

draft-authors-mole-crypto.md
```diff
@@ -684,14 +686,14 @@ def Commit(skA, ctx_iss):
...
 
 The Anchor stores `state` for the duration of the session and sends `commitment`
-to the Client in a `CommitMessage` ({{wire}}). Note that `skA` is not used in
-`Commit`; it appears in the interface because an implementation MAY choose to
-carry it in the session state rather than reload it in `Respond`.
+to the Client in a `CommitMessage` ({{wire}}).
 
 `Commit` draws all of its randomness in one call and splits it into one seed
 per scalar ({{derive-scalar}}). A test vector fixes the single value `rand`;
 `y` is nonzero by construction.
```
cjpatton, line 695: How about just passing rand to Commit explicitly? Likewise for the others.

draft-authors-mole-crypto.md
```diff
@@ -991,24 +992,14 @@ particular, deserializing an `Element` rejects the group identity element.
...
-`body` is **empty**: it is a trigger, not a protocol message, and the three moves
-of {{scheme}} are executed after it. The "Exchanges" field of the IHAT
-registration in {{PROTOCOLS}} counts HTTP exchanges, of which there are still
-two. Neither context appears on the wire ({{context-binding}}).
-
-| Exchange | Request body | Response body |
-|---|---|---|
-| 1 | empty | `CommitMessage` |
-| 2 | `ChallengeMessage` | `ResponseMessage` |
-{: title="Mapping of the three issuance messages onto the two HTTP exchanges"}
```
cjpatton, line 1007: I don'/t think this table is particularly helpful.

draft-authors-mole-crypto.md
cjpatton, line 1008: Since the session identifier has security considerations, we should specify a fixed length for it rather than leaving this up to the Anchor.

draft-authors-mole-crypto.md
```diff
@@ -1164,18 +1144,25 @@ reconstruction of `A`:
...
-The Client therefore also proves knowledge of `delta` relating `X_hat` to a key
-in `anchor_set`, without revealing which key.
+Because this shift is *additive*, so issuance is left unmodified and the
+unforgeability of {{issuance}} carries over to the modified Endorsement
+({{security-considerations}}). But `X_hat` by itself provides no evidence that
+it was derived from an Anchor key. The Client therefore also proves knowledge
+of `delta` relating `X_hat` to a key in `anchor_set`, without revealing which
+key.
 
 ## The Issuer-Hiding Proof {#issuer-proof}
```
cjpatton, line 1154: Can we rename this to "redemption proof" to match Moussaka? It's also a bit clearer.

draft-authors-mole-crypto.md
```diff
@@ -1164,18 +1144,25 @@ reconstruction of `A`:
...
 
+> XXX Can we rename this to "redemption proof" to match Moussaka?
+
 The Client proves knowledge of `delta` such that
 `X_hat - anchor_set[i] = delta * B` for at least one `i`, without revealing
 `i`. This is a one-out-of-`n`
-disjunction of Schnorr statements, all of them over the same base `B`.
+disjunction of Schnorr statements, all of them over the same base
```
cjpatton, line 1159: Is there a way to reword this without saying "Schnorr statement"? We don't define this anywhere, so I don't think this will be clear to all readers.

draft-authors-mole-crypto.md
```diff
@@ -1240,8 +1227,11 @@ n = len(anchor_set)
 return (Y, q)
 ~~~
```
cjpatton, line 1223: GPT-5.6 Sol made the following comment here:

> The draft's range expression is malformed. This is the apparent intended transcription; unlike the preceding revision, it does not pad.

draft-authors-mole-crypto.md
```diff
@@ -1267,19 +1255,25 @@ be opened to any value. The Client commits the branch commitment of `index` at
 the binding position, and after the challenge is known it opens every other
 position to the branch commitment that the shared response determines. The
 branch at the binding position uses the actual witness and a commitment fixed
-before the challenge; which branch that is remains hidden by the commitment
+before the challenge; the index of the branch remains hidden by the commitment
 key.
 
 This document uses the commitment of the appendix of {{STACKSIG}} over pairs,
 and then assembles them into a tree. The inputs to the commitments are arbitrary
 strings of bytes, and the outputs are compressed group elements.
```
cjpatton, line 1256: Our encoding of P256 points is always compressed, correct? Do we need to reiterate here?

draft-authors-mole-crypto.md
```diff
@@ -1267,19 +1255,25 @@ be opened to any value. The Client commits the branch commitment of `index` at
...
+> XXX Our encoding of P256 points is always compressed, correct? Do we need to
+> reiterate here?
+
 We require a random permutation `P` on group elements. We defer discussing its
-instantiation to a later section. Being a permutation `P` has an inverse `Pinv`.
+instantiation to a later section. Being a permutation, `P` has an inverse,
+which we denote by `Pinv`.
 
 If we need a 1-of-1 binding commitment, we use the identity function. Thankfully
 we apply this commitment to group elements, so the output is always the same size.
```
cjpatton, line 1262: I don't understand what this means :)

draft-authors-mole-crypto.md
```diff
@@ -1253,11 +1243,9 @@ branches without revealing which branch produced it.
 It follows that a verifier given `proof_challenge` and one `response` can
 compute a commitment for every branch, and that every branch then verifies by
 construction. Nothing is proved by the responses themselves. What has to be
-enforced is that the Client fixed the commitment of one branch *before* the
+enforced is that the Client fixed the commitment of one branch before the
 challenge existed, and could not afterwards change it.
 
-
-
 ### Partially Binding Commitments {#pbvc}
```
cjpatton, line 1241: I didn't try all that hard, but I don't quite understand how batched Sigma works from this description. A slightly slower build up with a small running example, would probably help. An aasvg diagram would go along way as well.

draft-authors-mole-crypto.md
```diff
@@ -1361,11 +1357,20 @@ def ComputeProofChallenge(anchor_set, X_hat, endorsement, ctx_iss,
   return G.HashToScalar(proof_transcript)
 ~~~
```
cjpatton, line 1342: Why bind the issuance and redemption context? The redemption context is covered by the issued signature, so I don't see why it needs to be covered by the transcript. Likewise for the issuance context: If the context doesn't match the value used during issuance, then signature verification should fail.  Otherwise, there's nothing that guarantees that all three parties agree on the value.

draft-authors-mole-crypto.md
```diff
@@ -1440,28 +1420,6 @@ statement from the single response, equivocates each level to the value its
...
-and with it `X_hat - delta * B`, which is the public key of the Anchor that
-issued the Endorsement. Reuse therefore does not merely weaken issuer hiding, it
-destroys it, and it strips the rerandomization from the signature for everyone,
-not only for the Moderator. A repeated commitment key is in any case a value a
-verifier can recognize, and would link the two redemptions to each other.
-
-Proving is linear in `n`: the second loop computes `N` branch commitments and
-`N - 1` node commitments, however short the proof is. The saving is in
-communication, not in computation; see {{security-considerations}}.
-
```
cjpatton, line 1464: The first paragraph has a pretty gross Claudism in it. But stepping back, this section feels a lot like Claude convincing itself that something works. I don't think it's a super useful analysis in the final product.

I was also wondering what `N` is. I'm not sure if that's part of an AI's context that's missing in the final product.

draft-authors-mole-crypto.md
```diff
@@ -1609,19 +1567,13 @@ def Redeem(anchor_set, index, endorsement, ctx_iss, ctx_red,
...
-`Redeem` raises a `VerifyError` rather than producing a redemption. This is the
-Client's own protection, and it is not redundant with the requirement on the
-Moderator not to offer such a set ({{verify-redemption}}): that requirement is
-of no help against a Moderator that offers a single key precisely in order to
-learn which Anchor endorsed this Client. Refusing also keeps the depth `q` of
-the tree at least one, so a redemption always carries at least one commitment
-key.
+In order to keep the Anchor anonymous, the Client MUST NOT redeem against an
+Anchor Set of fewer than two keys. Refusing also keeps the depth `q` of the
+tree at least one, so a redemption always carries at least one commitment key.
```
cjpatton, line 1572: This requirement is fine, but we will likely live in a world for a little while in which there is only one Anchor, and we want to redeem Endorsements anyway. How will this work?

draft-authors-mole-crypto.md
```diff
@@ -1609,19 +1567,13 @@ def Redeem(anchor_set, index, endorsement, ctx_iss, ctx_red,
...
 
-The amount of randomness a redemption consumes depends on the size of the Anchor
-Set, through the depth `q` of the tree ({{pbvc}}), and on nothing else; in
-particular it does not depend on `index`. `Statements` is called here only to
-obtain `q`, which is why the rerandomized key it is given is irrelevant.
```
cjpatton, line 1624: This seems unnecessary.

draft-authors-mole-crypto.md
```diff
@@ -1727,6 +1679,8 @@ the Moderator offered ({{pbvc}}). The two openings of level `j` are at positions
 other length with a `DeserializeError` rather than truncating or padding, and
 MUST NOT infer the size of its own Anchor Set from the redemption.
```
cjpatton, line 1676: Give an explicit formula for `q`.

draft-authors-mole-crypto.md
```diff
@@ -1736,8 +1690,9 @@ and a redemption is 649 bytes, of which 456 are the proof; at 64 keys it is
 smaller of the two for every Anchor Set of six or more keys, and its advantage
 grows with the set, which is the reason for using it.
```
cjpatton, line 1685: I'm guessing the estimates for disjunction is generated by AI. Instead I think it ought to be generated by a reference implemntation. Absent this implementation, I think we should not include these numbers.

draft-authors-mole-crypto.md
```diff
@@ -1821,56 +1766,6 @@ P:
...
-  set to zero; `Ns = 32`.
-
-DeserializeScalar(buf):
-: Deserialize a `Scalar` from a little-endian 32-byte string. Raise a
-  `DeserializeError` if the result is not in `[0, Order()-1]`; note that this
-  requires the top three bits of the input to be zero.
-
-P
-: Very similar to above, but 32 byte random permutations are easier to use instantiate
-with a symmetric Feisel network.
```
cjpatton, line 1872: This is likely to code rot. I say we remove it and add it back once someone asks for it.

draft-authors-mole-crypto.md
```diff
@@ -1883,13 +1778,6 @@ the Anchor's session state, and the seed drawn in `Challenge` determines the
 Client's blinding factors, so recovering either undoes the property that
 algorithm provides.
 
-> **Editorial note.** This document does not use the `RandomScalar()` member of
-> {{OPRF}}'s group interface, and so does not have to resolve an inconsistency
-> in it: {{Section 2.1 of OPRF}} defines `RandomScalar()` as "Chooses at random
-> a nonzero element in GF(p)", while each of its ciphersuites defines it as "a
-> uniformly random Scalar in the range `[0, G.Order() - 1]`", which includes
-> zero. **TODO:** report the inconsistency against {{OPRF}}.
-
```
cjpatton, line 1892: https://github.com/cfrg/draft-irtf-cfrg-voprf/issues/392

draft-authors-mole-crypto.md
```diff
@@ -1883,13 +1778,6 @@ the Anchor's session state, and the seed drawn in `Challenge` determines the
...
 # Security Considerations {#security-considerations}
```
cjpatton, line 1781: Note: I haven't looked closely at this section yet.

draft-authors-mole-crypto.md
wbl: Information theoretic privacy wouldn't hold, but that's not what we need: a PRF is fine to use everywhere.

draft-authors-mole-crypto.md
```diff
@@ -161,9 +161,9 @@ at issuance time to two contexts. These may be used to limit the validity scope
 of each Endorsement, i.e., when and for whom it may later be used.
 
 * The issuance context `ctx_iss` is agreed out of band among the Client, the
-  Anchors, and the Moderator. `ctx_iss` can encode, for instance, the time
-  period in which the Endorsement is issued, allowing us to capture Endorsement
-  expiry.
+  Anchor, and the Moderator. `ctx_iss` can encode, for instance, the time
+  period in which the Endorsement is issued, ensuring Endorsements expire.
```
wbl, line 165: I think there's some question about where this should be specified: at a higher level protocol wrapping these crypto bits would be my guess. The wording change is fine by me.

draft-authors-mole-crypto.md
```diff
@@ -557,7 +559,7 @@ trusted to choose each.
 The examples below illustrate their cryptographic roles and do not define
 alternative context encodings.
 
-The issuance context, written `ctx_iss`, restricts *when* an Endorsement may be
+The issuance context, written `ctx_iss`, restricts when an Endorsement may be
```
wbl, line 562: When, and potentially where. Agree with the typography change

draft-authors-mole-crypto.md
```diff
@@ -604,7 +606,7 @@ context rather than per Client.
 
 The nullifier `nf` MUST be a fresh string of `Nn` uniformly random bytes,
 generated by the Client, and MUST NOT be reused across Endorsements. It is
-revealed at redemption, where the Moderator uses it to enforce that each
+revealed at redemption time, where the Moderator uses it to enforce that each
```
wbl, line 609:
```suggestion
revealed during redemption, so the Moderator ensures each
```

draft-authors-mole-crypto.md
wbl: I think this is potentially a much better way of proceeding, and could avoid the need to do a bunch of arithmetic up front as well as harden the test vectors against innocent reorderings. But we'd need to ensure each call using the seed was distinguished: not that hard, does slightly complicate loops but not that much. We'd then have GetRandomScalar(where, iter, rand) where iter can be used for loops and where has to be distinct for each callsite in the doc.

draft-authors-mole-crypto.md
```diff
@@ -891,16 +898,12 @@ def Finalize(pkA, state, response):
...
-nullifier; its encoding is given in {{endorsement-encoding}}. The contexts it
-is bound to are not part of it. They are inputs to verification, supplied by
-the verifier, and a Client keeps its own copy of them for as long as it holds
-the Endorsement.
-
-The Endorsement is held privately by the Client until it is redeemed
-({{redemption}}). The Anchor never sees it.
+nullifier; its encoding is given in {{endorsement-encoding}}. The Endorsement
+is held privately by the Client until it is redeemed ({{redemption}}). The
+Anchor never sees it.
```
wbl, line 906: The Anchor could be a Moderator, I'm not sure why this last sentence is needed. The common context should also be part of the Endorsement as stored as its needed for redemption.

draft-authors-mole-crypto.md
```diff
@@ -991,24 +994,14 @@ particular, deserializing an `Element` rejects the group identity element.
...
 `EndorsementRequest` and `EndorsementResponse` bodies defined in {{PROTOCOLS}},
-whose transport is HTTP.
-
-Issuance has three messages, but the Anchor sends the first of them, and HTTP is
-client-initiated. The Client therefore opens the session with a request whose
-`body` is **empty**: it is a trigger, not a protocol message, and the three moves
-of {{scheme}} are executed after it. The "Exchanges" field of the IHAT
-registration in {{PROTOCOLS}} counts HTTP exchanges, of which there are still
-two. Neither context appears on the wire ({{context-binding}}).
+whose transport is HTTP. The first request is empty, and the corresponding
```
wbl, line 997: @martinthomson who knows HTTP language better than I do: should we say empty request, or empty body?

wbl: Also if the common data that goes into Z is going to be client specified or anchor specified it has to go to the other party. That data will change if we use it to enforce expiry dates!

draft-authors-mole-crypto.md
wbl: So in chat we thought that maybe we can drop the session ID as the client and Anchor probably have a cookie. I think that's a bad idea now given that we're in a context where there might be concurrency and I don't think there are a lot of alternatives. That said, I'm not quite sure what security considerations apply uniformly here. There is a client Anchor authentication underlying this and the Anchor has a variety of sensible means to issue session_ids and use them that may vary with underlying deployment technologies. It could be contextual to the user identity in a cookie, or it could be some global storage address with some imposed structure from that system.

Overall this has a very non-RESTy flavor and we might find ourselves regretting that: this is especially the case for the session_id being represented this custom way, vs. in a place where intermediaries are frequently configured to poke at it.

draft-authors-mole-crypto.md
```diff
@@ -1037,25 +1030,12 @@ struct {
...
-### Session Identifier {#session-id}
-
-The Anchor holds secret state between its two moves ({{sessions}}), so the two
-exchanges have to be correlated. `session_id` does that. It is generated by
-the Anchor, opaque to the Client, and echoed unmodified in `ChallengeMessage`.
-
 An Anchor MUST NOT have two open sessions with the same `session_id`, and SHOULD
 generate it with a cryptographically secure random number generator so that a
 Client cannot guess, and so collide with, another Client's session. An Anchor
 that receives a `ChallengeMessage` whose `session_id` does not correspond to one
```
wbl, line 1036: I could see implementations being hamstrung by this language in cases where a single Anchor is geographically distributed and users are using only one instance of it.

draft-authors-mole-crypto.md
wbl: Could `one-out-of-n discrete logarithms` work?

draft-authors-mole-crypto.md
wbl: How as in why, or as in some of the details? The solutions are a bit different.

Yes, I think a few trees, particularly with leaves getting joined in at different heights due to odd size could really help. Will make a note to draw them.

draft-authors-mole-crypto.md
wbl: So when n=1, there's a great way to prove a 1-out-of-1 statement: just do it. But if the first message of that sigma protocol was a different size than the output of the commitment scheme, upper layer code would have to be deeply aware of that fact when serializing. Because that's not the case, we can just use the identity function as a 1-out-of-1 binding commitment without complications about the size of the returned bytes.

draft-authors-mole-crypto.md
wbl: I think in that case you don't get any anonymity, and letting the commitment scheme be the identity works.

draft-authors-mole-crypto.md
wbl, line 1672: This is now wrong

draft-authors-mole-crypto.md
wbl: 2^ceil(log_2(n)) I think.

wbl (approved):

draft-authors-mole-crypto.md
cjpatton: The opinion of the draft currently is that it should be specified at the protocol layer. Personally I think that's a much cleaner design. I'll add a TODO to track.

draft-authors-mole-crypto.md
cjpatton: Agreed. I'll add as a TODO here.

draft-authors-mole-crypto.md
cjpatton: tracking here: https://github.com/Moderation-of-unLinkable-Endorsements/internet-drafts/issues/72

draft-authors-mole-crypto.md
cjpatton: Ok, I'm gonna resolve this one but make a note to bring it up for discussion next time we meet.

draft-authors-mole-crypto.md
cjpatton: Done.

draft-authors-mole-crypto.md
cjpatton: making a note to discuss this one at the next PACT sync.

draft-authors-mole-crypto.md
cjpatton: Deleted this paragraph.

draft-authors-mole-crypto.md
cjpatton: Agreed. I removed the last sentence.

draft-authors-mole-crypto.md
cjpatton: Removed the paragraph.

PR #70, opened 2026-09-14
give IHAT its own draft

SamuelSchlesinger:
this renames the crypto draft to IHAT and updates the references. the protocol itself is unchanged. ACT will live in a separate draft.

ACT follows in #67.


cjpatton (approved):
Boom!

dennisjackson (approved):

wbl:
We should probably add the name Rollatini to differentiate construction from functionality somewhere

